PR #2, opened 2020-05-14
Implement controller to bootstrap a vanilla cluster with no replicas.

tapih:
Implement [this task](https://app.zenhub.com/workspaces/neco-5ca28d055128c041a68c0a8f/issues/cybozu-go/neco/956)

Signed-off-by: Hiroshi Muraoka <h.muraoka714@gmail.com>


cmd/entrypoint/cmd/init.go
```diff
@@ -0,0 +1,390 @@
...
+	err = initializeOperatorAdminUser(ctx, viper.GetString(myso.OperatorPasswordFlag))
+	if err != nil {
+		return err
+	}
+
+	log.Info("setup ping user", nil)
+	err = initializePingUser(ctx)
+	if err != nil {
+		return err
+	}
```
ymmt2005, line 124: This user may mistakenly determine the mysql process is not ready or running
because it does not have any privileges.

cmd/entrypoint/cmd/init.go
```diff
@@ -0,0 +1,390 @@
...
+
+	cmd := well.CommandContext(ctx, "mysqld", "--skip-networking", "--socket="+socketPath)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	return cmd.Start()
+}
+
+func waitInstanceBootstrap(ctx context.Context) error {
+	timeout := time.After(timeoutSeconds * time.Second)
+	tick := time.Tick(time.Second)
```
ymmt2005, line 162: `time.Tick` leaks.  Use `time.NewTicker` and call `Stop()`.
https://pkg.go.dev/time?tab=doc#Ticker.Stop

cmd/entrypoint/cmd/init.go
```diff
@@ -0,0 +1,390 @@
...
+		case <-timeout:
+			return errors.New("timed out")
+		case <-tick:
+			_, _, err := doExec(ctx, nil, "mysqladmin", "--socket="+socketPath, "ping")
+			if err == nil {
+				return nil
+			}
+		}
+	}
+}
```
ymmt2005, line 174:
```suggestion
func waitInstanceBootstrap(ctx context.Context) error {
	ctx, cancel := context.WithTimeout(ctx, timeoutDuration)
	defer cancel()
	tick := time.NewTicker(time.Second)
	defer tick.Stop()

	for {
		select {
		case <-ctx.Done():
			return ctx.Err()
		case <-tick:
			_, _, err := doExec(ctx, nil, "mysqladmin", "--socket="+socketPath, "ping")
			if err == nil {
				return nil
			}
		}
	}
}
```

cmd/entrypoint/cmd/init.go
```diff
@@ -0,0 +1,390 @@
...
+func doExec(ctx context.Context, input []byte, command string, args ...string) ([]byte, []byte, error) {
+	var stdout, stderr bytes.Buffer
+	cmd := well.CommandContext(ctx, command, args...)
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+	if input != nil {
+		cmd.Stdin = bytes.NewReader(input)
+	}
+
+	err := cmd.Run()
```
ymmt2005, line 363: Use `cmd.Output()` for this purpose.

cmd/entrypoint/cmd/root.go
```diff
@@ -0,0 +1,33 @@
...
+
+// Execute executes the root command.
+func Execute() error {
+	return rootCmd.Execute()
+}
+
+func init() {
+	cobra.OnInitialize(initConfig)
+
+	rootCmd.AddCommand(initCmd)
```
ymmt2005, line 26: This should be written in `init.go`.

controllers/mysqlcluster_controller.go
```diff
@@ -0,0 +1,426 @@
...
+	}
+	return nil
+}
+
+func (r *MySQLClusterReconciler) makePodTemplate(log logr.Logger, cluster *mysov1alpha1.MySQLCluster) corev1.PodTemplateSpec {
+	template := cluster.Spec.PodTemplate
+	newTemplate := corev1.PodTemplateSpec{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        template.Name,
+			Namespace:   cluster.Namespace,
```
ymmt2005, line 245: No need.

Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+COPY cmd/ cmd/
+COPY constants.go constants.go
+COPY version.go version.go
+
+# Build
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o moco-controller ./cmd/moco-controller/main.go
+
+# Use distroless as minimal base image to package the moco-controller binary
+# Refer to https://github.com/GoogleContainerTools/distroless for more details
+FROM gcr.io/distroless/static:nonroot
```
ymmt2005, line 23: For what reason do you want to use distroless?

.circleci/config.yml
```diff
@@ -1,15 +1,90 @@
...
+      - e2e-command
 
 workflows:
   version: 2
   main:
     jobs:
       - build
+      - image:
+          requires:
+            - build
```
ymmt2005, line 89: Dockerfile seems not to use the built binary, doesn't it?

Makefile
```diff
@@ -1,26 +1,86 @@
...
+	test -z "$$(go vet ./... | grep -v '^vendor' | tee /dev/stderr)"
+
+# Build moco-controller binary
+moco-controller: generate
+	go build -o bin/moco-controller ./cmd/moco-controller/main.go
+
+# Generate manifests e.g. CRD, RBAC etc.
+manifests: controller-gen
+	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=moco-controller-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+	rm -f deploy/crd.yaml
```
ymmt2005, line 47: seems unnecessary.

Makefile
```diff
@@ -1,26 +1,86 @@
...
+# find or download controller-gen
+# download controller-gen if necessary
+controller-gen:
+ifeq (, $(shell which controller-gen))
+	@{ \
+	set -e ;\
+	CONTROLLER_GEN_TMP_DIR=$$(mktemp -d) ;\
+	cd $$CONTROLLER_GEN_TMP_DIR ;\
+	go mod init tmp ;\
+	go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.5 ;\
```
ymmt2005, line 63: Why don't you use `CTRLTOOLS_VERSION` as previously defined?
Or if installing controller-gen is a matter of `make setup`, remove this target altogether.

PROJECT
```diff
@@ -0,0 +1,7 @@
+domain: cybozu.com
+repo: github.com/cybozu-go/myso
```
ymmt2005, line 2: It seems to me renaming the repo now can reduce later troubles.

Dockerfile
binoue: This Dockerfile was originally generated by kube-builder.
Therefore, there is no good reason to use distroless and we will use our ubuntu image instead.

Makefile
morimoto-cybozu: Removed `controller-gen` target.

.circleci/config.yml
morimoto-cybozu: As you wrote, Dockerfile does not use any output of `build`.  `build` does not produce anything as persistent objects.
But I still want to put it in `requires` because `build` runs tests and it works as a requisite.  I also want to put `e2e` in the `requires` list.

controllers/mysqlcluster_controller.go
```diff
@@ -0,0 +1,476 @@
...
+		log.Info("reconcile StatefulSet successfully", "op", op)
+	}
+	return nil
+}
+
+func (r *MySQLClusterReconciler) makePodTemplate(log logr.Logger, cluster *mysov1alpha1.MySQLCluster) corev1.PodTemplateSpec {
+	template := cluster.Spec.PodTemplate
+	newTemplate := corev1.PodTemplateSpec{
+		ObjectMeta: metav1.ObjectMeta{
+			Labels:      template.Labels,
```
ymmt2005, line 255: You need to shallow copy `template.Labels` because it is later modified.

ymmt2005: Please check if there are similar problems.

controllers/mysqlcluster_controller.go
```diff
@@ -0,0 +1,476 @@
...
+			Annotations: template.Annotations,
+		},
+		Spec: template.Spec,
+	}
+
+	// add labels to describe application
+	if newTemplate.Labels == nil {
+		newTemplate.Labels = make(map[string]string)
+	}
```
ymmt2005, line 264:
```suggestion
	newTemplate := corev1.PodTemplateSpec{
		ObjectMeta: metav1.ObjectMeta{
			Annotations: template.Annotations,
		},
		Spec: template.Spec,
	}

	newTemplate.Labels = make(map[string]string)
	for k, v := range template.Labels {
		newTemplate.Labels[k] = v
	}
```

controllers/mysqlcluster_controller.go
binoue: > You need to shallow copy template.Labels because it is later modified.

This means that we need to use deep copy?

ymmt2005: @binoue 
No, shallow copy is correct.
Deep copy is recursive, while shallow copy is not.

I suggested a fix already.

api/v1alpha1/mysqlcluster_types.go
```diff
@@ -0,0 +1,242 @@
...
+type RestoreSpec struct {
+	// SourceClusterName is a source `MySQLCluster` name.
+	// +kubebuilder:validation:Required
+	SourceClusterName string `json:"restore"`
+
+	// PointInTime is a point-in-time of the state which the cluster is restored to.
+	// +kubebuilder:validation:Required
+	PointInTime metav1.Time `json:"pointInTime"`
+
+	// ObjectStorageName is a name of `ObjectStorage`.
```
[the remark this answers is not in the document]
ymmt2005: @binoue How about this?

PR #5, opened 2020-06-01
Add recovery from blackouts scenario to design doc.

morimoto-cybozu:

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -146,6 +146,21 @@ In this section, the name of `StatefulSet` is assumed to be `mysql`.
   - Remove `.spec.updateStrategy.rollingUpdate.partition`.
 6. Wait for `mysql-0` to be upgraded.
 
+### How to manage recovery from blackouts
+
+In the scenario of the recovery from data center blackouts, all members of the MySQL cluster perform cold boots.
+
+The operator waits for all members to boot up again.
+It automatically recovers the cluster only after all members come back, not just after the quorum come back.
+This is to prevent the data loss even in corner cases.
+
+If a part of the cluster never finishes boot-up, users must intervene the recovery process.
+1. Confirm that the quorum of the cluster are up and running, and waiting to be configured for the cluster.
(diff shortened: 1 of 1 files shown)


docs/design.md
```diff
@@ -146,6 +146,21 @@ In this section, the name of `StatefulSet` is assumed to be `mysql`.
...
+The operator waits for all members to boot up again.
+It automatically recovers the cluster only after all members come back, not just after the quorum come back.
+This is to prevent the data loss even in corner cases.
+
+If a part of the cluster never finishes boot-up, users must intervene the recovery process.
+1. Confirm that the quorum of the cluster are up and running, and waiting to be configured for the cluster.
+2. Stop the operator.
+3. Delete the non-running Pods and/or PVCs and/or Nodes, according to the level of the disruption.
+4. Configure the cluster with the remaining nodes.
+5. Resume the operator.
```
ymmt2005, line 162: It seems to me that 4 is very dangerous and too difficult for normal humans
because they could know almost nothing about how to configure clusters.

I'd like to suggest:
```suggestion
1. Delete the problematic Pods and/or PVCs.  (or they might have been deleted already)
2. StatefulSet controller creates new blank Pods and/or PVCs.
3. The operator tries to select the master of the cluster only after all members are up and running and the quorum of the cluster has data.
4. The operator initializes the new blank Pods as new slaves.
```

Step 1 is the key.  This step may happen unavoidably if the Kubernetes
cluster is backed by some automatic Node maintenance operators such as
CKE or [the cluster API](https://github.com/kubernetes-sigs/cluster-api).

toshipp: I think recovery operation through the operator is more safe option instead of interacting with kubernetes resources directly.
In this scenario, operator reports what node/pod is wrong to an administrator, then s/he submits a new CR for recovery or edits existing MySQLCluster.

Manual recovery is not avoidable when the operator can not handle the failure, but manual operation should be the last choice.

PR #39, opened 2020-06-24
Merge MySQL configuration ConfigMaps for moco-conf-gen.

morimoto-cybozu:


constants.go
```diff
@@ -43,12 +46,24 @@ const (
 	// PodNameFlag is a name of the flag of a pod name.
 	PodNameFlag = "pod-name"
 
+	// PodNamespaceEnvName is a name of the environment variable of a pod namespace.
+	PodNamespaceEnvName = "MYSQL_POD_NAMESPACE"
```
ymmt2005, line 50: There is a convention to name this as `POD_NAMESPACE`.
See https://github.com/kubernetes/kubernetes/blob/323f34858de18b862d43c40b2cced65ad8e24052/staging/src/k8s.io/client-go/tools/clientcmd/client_config.go#L527

controllers/mysqlcluster_controller.go
```diff
@@ -260,6 +296,101 @@ func generateRandomBytes(n int) ([]byte, error) {
 	return ret, nil
 }
 
+func mergeMycnf(cnf1, cnf2 map[string]string) {
```
ymmt2005, line 299: Please add comments to describe `cnf1` will be altered.

controllers/mysqlcluster_controller.go
```diff
@@ -260,6 +296,101 @@ func generateRandomBytes(n int) ([]byte, error) {
...
+	for k, v := range cnf2 {
+		cnf1[normalizeConfKey(k)] = v
+	}
+}
+
+func normalizeConfKey(k string) string {
+	return strings.ReplaceAll(k, "-", "_")
```
ymmt2005, line 306: I guessed the reverse because `consMycnf` defines keys using `-`.

controllers/mysqlcluster_controller.go
```diff
@@ -260,6 +296,101 @@ func generateRandomBytes(n int) ([]byte, error) {
...
+}
+
+func (r *MySQLClusterReconciler) createOrUpdateConfigMap(ctx context.Context, log logr.Logger, cluster *mocov1alpha1.MySQLCluster) error {
+	cm := &corev1.ConfigMap{}
+	cm.SetNamespace(cluster.Namespace)
+	cm.SetName(cluster.Name)
+
+	op, err := ctrl.CreateOrUpdate(ctx, r.Client, cm, func() error {
+		cm.Labels = map[string]string{
```
ymmt2005, line 315: Are you sure to drop all existing labels?

controllers/mysqlcluster_controller.go
```diff
@@ -260,6 +296,101 @@ func generateRandomBytes(n int) ([]byte, error) {
...
+			appNameKey: fmt.Sprintf("%s-%s", appName, cluster.Name),
+		}
+
+		conf := make(map[string]mysqlConf)
+		conf["mysqld"] = make(mysqlConf)
+		mergeMycnf(conf["mysqld"], defaultMycnf)
```
ymmt2005, line 321: This looks dangerous because of the necessities of `make` maps.
Why don't you abstract it further?

controllers/mysqlcluster_controller.go
```diff
@@ -260,6 +296,101 @@ func generateRandomBytes(n int) ([]byte, error) {
...
+
+		if cluster.Spec.MySQLConfigMapName != nil {
+			cm := &corev1.ConfigMap{}
+			err := r.Get(ctx, client.ObjectKey{Namespace: cluster.Namespace, Name: *cluster.Spec.MySQLConfigMapName}, cm)
+			if err != nil {
```
ymmt2005, line 326: Don't you need to check it with [`apierrors.IsNotFound`](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors?tab=doc#IsNotFound)?

controllers/mysqlcluster_controller.go
```diff
@@ -260,6 +296,101 @@ func generateRandomBytes(n int) ([]byte, error) {
...
+	cm := &corev1.ConfigMap{}
+	err := r.Get(ctx, objectKey, cm)
+	if err != nil {
+		return err
+	}
+	for k, v := range cm.Data {
+		conf[k] = v
+	}
+	return nil
+}
```
ymmt2005, line 366: What's this?

controllers/mysqlcluster_controller.go
```diff
@@ -260,6 +296,101 @@ func generateRandomBytes(n int) ([]byte, error) {
...
+	// sort keys to generate reproducible my.cnf
+	sections := make([]string, 0, len(conf))
+	for sec := range conf {
+		sections = append(sections, sec)
+	}
+	sort.Strings(sections)
+
+	b := new(strings.Builder)
+	for _, sec := range sections {
+		fmt.Fprintf(b, "[%s]\n", sec)
```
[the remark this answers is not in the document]
ymmt2005: Check `fmt.Frpintf` return values
https://golang.org/pkg/fmt/#Fprintf

controllers/mysql_conf.go
```diff
@@ -0,0 +1,83 @@
...
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+)
+
+var (
+	// Default options of mysqld section
+	defaultMycnf = map[string]string{}
```
ymmt2005, line 11: Please add these.
```
character-set-server = utf8mb4
collation-server = utf8mb4_unicode_ci
skip-character-set-client-handshake
default-time-zone = +0:00
```

ref. 
- https://stackoverflow.com/questions/3513773/change-mysql-default-character-set-to-utf-8-in-my-cnf
- https://github.dev.cybozu.co.jp/hazama/mysql/blob/master/cybozu/resources/etc/opt/cybozu/mysql/base.cnf

controllers/mysql_conf.go
```diff
@@ -0,0 +1,83 @@
...
+		"mysqld": {
+			"datadir":          "/var/lib/mysql",
+			"pid_file":         "/var/run/mysqld/mysqld.pid",
+			"socket":           "/var/run/mysqld/mysqld.sock",
+			"secure_file_priv": "NULL",
+
+			// Disabling symbolic-links to prevent assorted security risks
+			"symbolic_links": "0",
+			"server_id":      "{{ .server_id }}",
+			"admin_address":  "{{ .admin_address }}",
```
ymmt2005, line 23: Please add these.

```
skip-name-resolve
max_connections = 100000
```

controllers/mysqlcluster_controller.go
ysksuzuki: Yes, we're pretty sure, because Moco manages the cm, and there're no labels added other than that.

controllers/mysqlcluster_controller.go
```diff
@@ -260,6 +296,101 @@ func generateRandomBytes(n int) ([]byte, error) {
...
+}
+
+func (r *MySQLClusterReconciler) createOrUpdateConfigMap(ctx context.Context, log logr.Logger, cluster *mocov1alpha1.MySQLCluster) error {
+	cm := &corev1.ConfigMap{}
+	cm.SetNamespace(cluster.Namespace)
+	cm.SetName(cluster.Name)
+
+	op, err := ctrl.CreateOrUpdate(ctx, r.Client, cm, func() error {
+		cm.Labels = map[string]string{
+			appNameKey: fmt.Sprintf("%s-%s", appName, cluster.Name),
```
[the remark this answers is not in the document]
ysksuzuki: The label `app.kubernetes.io/name` describes both application and cluster name since argocd also uses label `app.kubernetes.io/instance`. So we added an application name to it.

controllers/mysqlcluster_controller.go
ysksuzuki: If a cm doesn't exist in a cluster namespace,  the controller only returns an error. Is there anything it would care about?

controllers/mysqlcluster_controller.go
ymmt2005: OK

controllers/mysqlcluster_controller.go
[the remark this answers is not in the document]
ymmt2005: @ysksuzuki 
So, what about `app.kubernetes.io/part-of` or `app.kubernetes.io/managed-by`?
https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels

controllers/mysqlcluster_controller.go
ymmt2005: Ah, okay.  only when `cluster.Spec.MySQLConfigMapName` is not nil.

controllers/mysqlcluster_controller.go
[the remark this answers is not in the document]
ymmt2005: Also I think `app.kubernetes.io/instance: moco` is ignored by Argo CD.
Argo CD cares about the label only when the label value is `argocd`, doesn't it?

controllers/mysql_conf.go
```diff
@@ -0,0 +1,178 @@
...
+		"tmpdir":                              moco.TmpPath,
+		"innodb_tmpdir":                       moco.TmpPath,
+		"character_set_server":                "utf8mb4",
+		"collation_server":                    "utf8mb4_unicode_ci",
+		"skip_character_set_client_handshake": "ON",
+
+		"default_time_zone": "+0:00",
+
+		"back_log":            "900",
+		"max_connections":     "5000",
```
ymmt2005, line 24: I think a larger value, for instance, `100000` is safer/better because 5,000 has no good reasons.

controllers/mysql_conf.go
```diff
@@ -0,0 +1,178 @@
...
+
+		// MyISAM Specific options
+		"key_buffer_size":           "32M",
+		"read_buffer_size":          "2M",
+		"read_rnd_buffer_size":      "4M",
+		"bulk_insert_buffer_size":   "8M",
+		"myisam_sort_buffer_size":   "4M",
+		"myisam_max_sort_file_size": "128M",
+		"myisam_repair_threads":     "1",
+		"myisam_recover_options":    "FORCE,BACKUP",
```
ymmt2005, line 59: MyISAM tables are not transactional and therefore is not
suitable for use in MOCO clusters.

Why don't you disable MyISAM altogether by `disabled_storage_engines="MyISAM"`?

controllers/mysql_conf.go
```diff
@@ -0,0 +1,197 @@
...
+		"innodb_log_file_size":      "800M",
+		"innodb_log_files_in_group": "2",
+		"innodb_read_io_threads":    "8",
+		"innodb_write_io_threads":   "4",
+		"innodb_io_capacity":        "400",
+		"innodb_io_capacity_max":    "3200",
+		"innodb_flushing_avg_loops": "10",
+		"innodb_purge_threads":      "1",
+		"innodb_page_cleaners":      "1",
+		"innodb_lru_scan_depth":     "256",
```
ymmt2005, line 76: Are these optimized for SSDs?  It seems to me, for example, `innodb_io_capacity` is too small for SSDs.

controllers/mysql_conf.go
```diff
@@ -0,0 +1,197 @@
...
+
+		"innodb_buffer_pool_in_core_file": "OFF", // It is rarely necessary to include a buffer pool in a core file.
+	}
+
+	constMycnf = map[string]map[string]string{
+		"mysqld": {
+			"port":                          "3306",
+			"socket":                        filepath.Join(moco.VarRunPath, "mysqld.sock"),
+			"datadir":                       moco.MySQLDataPath,
+			"default_authentication_plugin": "mysql_native_password",
```
ymmt2005, line 86: Why do you change this from the default, and why do you forbid users to modify this?

controllers/mysql_conf.go
```diff
@@ -0,0 +1,197 @@
...
+
+			"skip_name_resolve": "ON",
+
+			"log_error":           "/var/log/mysql/mysql.err",
+			"slow_query_log_file": "/var/log/mysql/mysql.slow",
```
ymmt2005, line 91: Do you inject sidecars to tail these files?

controllers/mysql_conf.go
```diff
@@ -0,0 +1,197 @@
...
+		"innodb_purge_threads":      "1",
+		"innodb_page_cleaners":      "1",
+		"innodb_lru_scan_depth":     "256",
+
+		"innodb_buffer_pool_in_core_file": "OFF", // It is rarely necessary to include a buffer pool in a core file.
+	}
+
+	constMycnf = map[string]map[string]string{
+		"mysqld": {
+			"port":                          "3306",
```
ymmt2005, line 83: Please disable MyISAM.
https://github.com/cybozu-go/moco/pull/39#discussion_r446146775

controllers/mysql_conf.go
ysksuzuki: Yes, we will handle it with the following task.

controllers/mysqlcluster_controller.go
```diff
@@ -186,12 +202,16 @@ func (r *MySQLClusterReconciler) createSecretIfNotExist(ctx context.Context, log
 }
 
 func (r *MySQLClusterReconciler) createPasswordSecretForUser(ctx context.Context, cluster *mocov1alpha1.MySQLCluster, rootPass, operatorPass, replicatorPass, donorPass []byte) error {
+	if cluster.Spec.RootPasswordSecretName != nil {
+		return nil
+	}
```
ymmt2005, line 207: as commented.

controllers/mysql_conf.go
```diff
@@ -0,0 +1,195 @@
...
+		"max_sp_recursion_depth": "20",
+
+		"print_identified_with_as_hex": "ON",
+
+		"loose_binlog_transaction_compression": "ON", // It would reduce the size of binlog by a third, but only available in 8.0.20 or later.
+
+		"information_schema_stats_expiry": "0", // No need to cache information_schema.tables values
+
+		// INNODB Specific options
+		"innodb_flush_method":              "O_DIRECT",
```
ymmt2005, line 61:
```
innodb_log_file_size = 800M
innodb_log_files_in_group = 2
```

docs/crd_mysql_cluster.md
```diff
@@ -22,11 +22,13 @@ MySQLClusterSpec
...
 | `serviceTemplate`             | [ServiceSpec]               | No       | `Service` template for both primary and replicas.                                                                                                                                 |
-| `mySQLConfigMapName`          | string                      | No       | `ConfigMap` name of MySQL config.                                                                                                                                                 |
+| `mysqlConfigMapName`          | string                      | No       | `ConfigMap` name of MySQL config.                                                                                                                                                 |
 | `rootPasswordSecretName`      | string                      | Yes      | `Secret` name for root user config.                                                                                                                                               |
```
ymmt2005, line 26: Please fix this as well.

controllers/mysql_conf.go
```diff
@@ -0,0 +1,198 @@
...
+		"innodb_lock_wait_timeout":         "60",
+		"innodb_print_all_deadlocks":       "1",
+		"innodb_online_alter_log_max_size": "1073741824",
+		"innodb_adaptive_hash_index":       "ON",
+		"innodb_numa_interleave":           "ON",
+		"innodb_buffer_pool_in_core_file":  "OFF", // It is rarely necessary to include a buffer pool in a core file.
+
+		// options for adaptive flushing
```
ymmt2005, line 71: These are not only for the adaptive flushing.  Remove the comment and a blank line.

```suggestion
```

controllers/mysql_conf.go
```diff
@@ -0,0 +1,196 @@
...
+		"disabled_storage_engines": "MyISAM",
+
+		// INNODB Specific options
+		"innodb_flush_method":              "O_DIRECT",
+		"innodb_lock_wait_timeout":         "60",
+		"innodb_print_all_deadlocks":       "1",
+		"innodb_online_alter_log_max_size": "1073741824",
+		"innodb_adaptive_hash_index":       "ON",
+		"innodb_numa_interleave":           "ON",
+		"innodb_buffer_pool_in_core_file":  "OFF", // It is rarely necessary to include a buffer pool in a core file.
```
ymmt2005, line 69: Please add these.
```
innodb_buffer_pool_dump_pct = 100
innodb_buffer_pool_dump_at_shutdown = 1
innodb_buffer_pool_load_at_startup = 0
```
https://tech.gmo-media.jp/post/128523146249/innodb-buffer-pool-load-at-startup

controllers/mysql_conf.go
ymmt2005: committed.

controllers/mysqlcluster_controller.go
```diff
@@ -260,18 +291,64 @@ func generateRandomBytes(n int) ([]byte, error) {
...
 	headless.SetNamespace(cluster.Namespace)
-	headless.SetName(cluster.Name)
+	headless.SetName(uniqueName(cluster))
 
 	op, err := ctrl.CreateOrUpdate(ctx, r.Client, headless, func() error {
 		headless.Labels = map[string]string{
-			appNameKey: fmt.Sprintf("%s-%s", appName, cluster.Name),
+			appNameKey:      uniqueName(cluster),
+			appManagedByKey: myName,
 		}
```
ymmt2005, line 347: This pattern appears too often.  How about creating a utility function like this:

```go
func setLabels(om metav1.ObjectMeta) {
    om.Labels = map[string]string{
        appNameKey: om.Name,
        appManagedByKey: myName,
    }
}
```

ref: https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1?tab=doc#ObjectMeta

controllers/mysqlcluster_controller.go
```diff
@@ -260,18 +290,67 @@ func generateRandomBytes(n int) ([]byte, error) {
...
+			log: log,
+		}
+		gen.mergeSection("mysqld", defaultMycnf, false)
+
+		// Set innodb_buffer_pool_size if resources.requests.memory or resources.limits.memory is specified
+		mem := getMysqldContainerRequests(cluster, corev1.ResourceMemory)
+		if mem != nil {
+			bufferSize := int64(float64(mem.Value()) * 0.7)
+			gen.mergeSection("mysqld", map[string]string{"innodb_buffer_pool_size": strconv.FormatInt(bufferSize, 10)}, false)
+		}
```
ymmt2005, line 310: I think it is better to round up the value in MiB.  How about this?
```suggestion
		// 128MiB is the default innodb_buffer_pool_size value
		if mem != nil && mem.Value() > (128<<20) {
			bufferSize := mem.Value() / 10 * 7
			gen.mergeSection("mysqld", map[string]string{"innodb_buffer_pool_size": fmt.Sprintf("%dM", bufferSize>>20), false)
		}
```

PR #43, opened 2020-07-08
Add periodic log rotation mechanism

zoetrope:


cmd/entrypoint/cmd/init.go
```diff
@@ -339,6 +345,7 @@ func init() {
 	initCmd.Flags().String(moco.PodIPFlag, "", "Pod IP address")
 	initCmd.Flags().String(moco.OperatorPasswordFlag, "", "Password for both operator user and operator admin user")
 	initCmd.Flags().String(moco.RootPasswordFlag, "", "Password for root user")
+	initCmd.Flags().String(moco.MiscPasswordFlag, "", "Password for misc user")
```
ymmt2005, line 348: Please tell me why you are taking passwords via command-line flags?
I'm afraid that these passwords can be leaked easily by just seeing the pod manifest.

cmd/entrypoint/cmd/rotate_server.go
```diff
@@ -0,0 +1,111 @@
...
+		return fmt.Errorf("failed to stat slow query log file: %w", err)
+	}
+
+	cmd := well.CommandContext(r.Context(), "mysql", "--defaults-extra-file="+filepath.Join(moco.MySQLDataPath, "misc.cnf"))
+	cmd.Stdin = strings.NewReader("FLUSH ERROR LOGS;\nFLUSH SLOW LOGS;\n")
+	err = cmd.Run()
+	if err != nil {
+		log.Error("failed to exec mysql FLUSH", map[string]interface{}{
+			log.FnError: err,
+		})
```
ymmt2005, line 95: `well.CommandContext` automatically does this.

docs/design.md
```diff
@@ -227,6 +227,17 @@ and export them as Prometheus metrics.
...
+### How to manage log files
+
+The operator configures MySQL to output error logs and slow query logs into files.
+
+To avoid exhaustion of storage resources, the operator appends a sidecar container to the MySQL Pod.
+The sidecar container rotates and deletes the log files.
+The operator creates a CronJob to invoke rotation and deletion periodically.
+
+The operator does not care about gathering the contents of the log files.
+Tenant users can extract the contents by defining a sidecar container.
```
ymmt2005, line 239:
```suggestion
Tenant users can extract the contents by defining sidecar containers for `slow.log` and `error.log`.
```

Could you add some examples?

api/v1alpha1/mysqlcluster_types.go
```diff
@@ -47,6 +47,11 @@ type MySQLClusterSpec struct {
 	// +optional
 	ReplicationSourceSecretName *string `json:"replicationSourceSecretName,omitempty"`
 
+	// LogRotationSchedule is a schedule in Cron format for MySQL log rotation
+	// +kubebuilder:default="*/5 * * * *"
```
ymmt2005, line 51: How do you validate the given value as a valid schedule?

controllers/mysqlcluster_controller.go
```diff
@@ -216,6 +227,10 @@ func (r *MySQLClusterReconciler) createPasswordSecretForUser(ctx context.Context
 			return err
 		}
 	}
+	miscPass, err := generateRandomBytes(passwordBytes)
```
ymmt2005, line 230: This is not a password for _users_.

ysksuzuki: Yes, you're right. A root password is the only one that worth checking by users. However, all passwords in the Secret are referred by the entrypoint when it creates MySQL users during the initializing process. So the Secret contains those passwords besides a root password. It's probably better to change the function name to a more appropriate one.

ymmt2005: How about just moving this to `r.createPasswordSecretForController`?
Are there some obstacles?

docs/design.md
ymmt2005: I think an example using filebeat is good.

controllers/mysqlcluster_controller.go
ysksuzuki: Currently, we cannot move those passwords to `r.createPasswordSecretForController`, because all passwords in the Secret are referred by the entrypoint when it creates MySQL users during the initializing process.
The entrypoint cannot refer to the Secret in other namespace.

ymmt2005: I see.  So, how about renaming this method to something like `createPasswordSecretForInit`
because the Secret is used primarily for initialization.

cmd/entrypoint/cmd/init.go
ysksuzuki: Actually we pass the passwords with the environment variables and don't use those command-line arguments. It's better to remove them.

Just to be clear, is it OK to pass passwords with the env variables from Secret?

api/v1alpha1/mysqlcluster_types.go
ysksuzuki: How about we validate it just like k8s does?
https://github.com/kubernetes/kubernetes/blob/b33e202397319fd85e5576b35c815e8ae2f76d7e/pkg/apis/batch/validation/validation.go#L241

k8s does it with [this library](https://github.com/robfig/cron).
In this case, we validate it with Admission Webhook.

Or we validate it with regex. In this case we can do it with `kubebuilder:validation:Pattern`.

What do you think?

cmd/entrypoint/cmd/init.go
ymmt2005: > Just to be clear, is it OK to pass passwords with the env variables from Secret?

It's okay because environment variables of others' processes can be viewed
only by the superuser.

api/v1alpha1/mysqlcluster_types.go
ymmt2005: Either of them is okay.

ysksuzuki: Alright, we go for the regex.

PR #78, opened 2020-10-13
use column alias name for expression query

umezawatakeshi:


kfyharukz (approved):
LGTM

PR #89, opened 2020-10-28
fix broken list in document.

umezawatakeshi:
fix broken list in document.


kfyharukz (approved):
LGTM

PR #101, opened 2020-11-04
fix primary failover test

umezawatakeshi:


umezawatakeshi:
This fix is cherry-picked into #105

PR #107, opened 2020-11-06
fix deployment manifest

umezawatakeshi:


morimoto-cybozu (approved):
mobbed

PR #109, opened 2020-11-10
Fix kubectl moco to enable interactive exec.

kfyharukz:
This PR fixes `kubectl-moco` to use the existing `kubectl exec` implements because it has more detailed error handling.


morimoto-cybozu (approved):
LGTM

PR #115, opened 2020-11-11
Fix resource requests/limits.

kfyharukz:
Signed-off-by: Kazuhito MATSUDA <kazuhito.matuda@gmail.com>

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -26,12 +26,8 @@ spec:
         - /moco-controller
         image: ghcr.io/cybozu-go/moco:latest
         resources:
-          limits:
-            cpu: 4000m
-            memory: 2Gi
           requests:
-            cpu: 1000m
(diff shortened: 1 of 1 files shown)


config/manager/manager.yaml
```diff
@@ -25,13 +25,6 @@ spec:
         command:
         - /moco-controller
         image: ghcr.io/cybozu-go/moco:latest
-        resources:
-          limits:
-            cpu: 4000m
-            memory: 2Gi
-          requests:
```
ymmt2005, line 32: Pods w/o resource requests have the lowest QoS priority (BestEffort) and should be avoided in general.
https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/

I'd suggest having `resources.requests.memory=256Mi`

ymmt2005 (approved):
LGTM

PR #124, opened 2020-12-10
docs: Add example of MySQLCluster CR.

morimoto-cybozu:


masa213f (approved):
LGTM

PR #131, opened 2020-12-22
Delete unused port definition

masa213f:
MOCO does not implement webhook, so the `ctrl.Options.Port` option is not necessary.


yokaze (approved):
LGTM

PR #134, opened 2020-12-22
Fix controller log additional field

masa213f:
- Remove duplicate output of MySQLCluster name. The name is set commonly. https://github.com/cybozu-go/moco/blob/54ad80e809100fa01c734dc6edcce7648c1771ac/controllers/mysqlcluster_controller.go#L98
- Output the operation (`CREATE` or `UPDATE`) when reconciling services.
- Lowercase the start of some log messages.
- Output the cluster status when the status could not be updated.


yokaze (approved):
LGTM

PR #135, opened 2020-12-23
Output logs on CI

masa213f:
Print logs of the following components.
- moco-controller-manager
- MySQLCluster in `e2e-test-external` ns

In this PR, I cannot print the log of MySQLCluster in`e2e-test` ns.
Because the resource is deleted in the k2e test.
https://github.com/cybozu-go/moco/blob/v0.3.1/e2e/gc_test.go


yokaze (approved):
LGTM

PR #138, opened 2020-12-25
Add comment for error check after cloning

masa213f:
Add a comment for the following behavior after cloning.

https://dev.mysql.com/doc/refman/8.0/en/clone-plugin-remote.html

>By default, the recipient MySQL server instance is restarted (stopped and started) automatically after the data is cloned. For an automatic restart to occur, a monitoring process must be available on the recipient to detect server shutdowns. Otherwise, the cloning operation halts with the following error after the data is cloned, and the recipient MySQL server instance is shut down:
>> ERROR 3707 (HY000): Restart server failed (mysqld is not managed by supervisor process).


yokaze (approved):
LGTM

PR #142, opened 2021-01-07
Run small test with MySQL 8.0.18 and 8.0.20

masa213f:
- Run the small test with MySQL 8.0.18 and 8.0.20
- Remove redundant or unused constants.
- Change the test password: `test-password` -> `testpassword`.
  - Because the GitHub Actions will mask the strings `-password <HOGE>` unexpectedly.
  - If not change the test password, this command is ...
     - `docker run <other param> -e MYSQL_ROOT_PASSWORD=test-password mysql:8.0.18`
  - logged as follows. 😱
     - `docker run <other param> -e MYSQL_ROOT_PASSWORD=test***`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ endif
 
 KUBEBUILDER_VERSION := 2.3.1
 CTRLTOOLS_VERSION := 0.4.0
-MYSQL_VERSION := 8.0.21
+MYSQL_VERSION ?= 8.0.20
 
 .PHONY: all
 all: build/moco-controller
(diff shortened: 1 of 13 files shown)


Makefile
```diff
@@ -41,24 +41,24 @@ validate: setup
 # Run tests
 .PHONY: test
 test: $(KUBEBUILDER)
-	go test -race -v -coverprofile cover.out ./...
+	go test -ldflags="-X github.com/cybozu-go/moco/test_utils.MySQLVersion=$(MYSQL_VERSION)" -race -v -coverprofile cover.out ./...
```
yokaze, line 44: I think it is better to use environment variables in this case.

agent/test_data/PRIMARY_PASSWORD
```diff
@@ -1 +1 @@
-test-password
\ No newline at end of file
+rootpassword
```
yokaze, line 1: It looks better to use different test passwords for different purposes.
Can we avoid the possibility to confuse the passwords in this directory?

yokaze (changes requested):

Makefile
zoetrope, line 44: I think you can use environment variable instead of embedding variable.
Which do you think is better?

agent/test_data/PRIMARY_PASSWORD
masa213f: It's good. However, it needs a lot of changes.
So I will do it in another PR.

https://github.com/cybozu-go/moco/pull/147

yokaze (approved):
LGTM

PR #143, opened 2021-01-08
Prevent the crash of the agent process when with the wrong external donor settings

masa213f:
The agent crashes with the wrong external donor settings by the combination of the following.

1. When the external donor setting is wrong, the agent cannot detect clone failure.
Then the agent will wait for bootstrap, but it will get an error because of timeout.
https://github.com/cybozu-go/moco/blob/v0.3.1/agent/clone.go#L168-L170

2. When the function in [this `well.Go`](https://github.com/cybozu-go/moco/blob/v0.3.1/agent/clone.go#L166-L199) is failed, [this `well.HTTPServer`](https://github.com/cybozu-go/moco/blob/v0.3.1/cmd/entrypoint/cmd/agent.go#L86-L97) will stop.
And the agent process will be finished.


agent/test_data/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
+!.gitignore
```
zoetrope, line 2: I think it would be better to put the following `.gitignore` in the agent directory.

```
/test_data
```

Then, create `test_data` directory in BeforeSuite.

zoetrope (commented):
LGTM

agent/clone_test.go
```diff
@@ -230,7 +242,139 @@ func testClone() {
...
+			}, 30*time.Second).Should(Succeed())
+		}
+
+		By("checking metrics")
+		Eventually(func() error {
+			cloneCount, err := getMetric(registry, metricsPrefix+"clone_count")
+			if err != nil {
+				return err
+			}
+			if *cloneCount.Counter.Value != 4.0 {
```
yokaze, line 340: It would be a bit informative if `4.0` is replaced with `len(testcases)`?

agent/clone_test.go
```diff
@@ -230,7 +242,139 @@ func testClone() {
...
+				return fmt.Errorf("clone_count isn't incremented yet: value=%f", *cloneCount.Counter.Value)
+			}
+
+			cloneFailureCount, err := getMetric(registry, metricsPrefix+"clone_failure_count")
+			if err != nil {
+				return err
+			}
+			if *cloneFailureCount.Counter.Value != 4.0 {
```
yokaze, line 348: ditto

yokaze (changes requested):

zoetrope (approved):
LGTM

yokaze (approved):
LGTM

PR #149, opened 2021-01-19
Fix controller test

masa213f:
Change the controllers test as follows.

- Simplify the accessor mock.
   This test does not seem to access a MySQL server. The parameters in the mock are confusing.
- Change the variable name. (`namespace` -> `clusterNamespace`)


yokaze (approved):
LGTM

zoetrope (approved):

PR #153, opened 2021-01-20
Update Go to 1.15 and Ubuntu to 20.04

masa213f:
- Update Go to 1.15
- Update Ubuntu to 20.04

`go install` outside `GOPATH` will fail with go 1.15 as follows. So I fixed the relevant part to use `go get`.
```
$ make custom-checker
if ! which custom-checker >/dev/null; then \
        cd /tmp; env GOFLAGS= GO111MODULE=on go install github.com/cybozu/neco-containers/golang/analyzer/cmd/custom-checker; \
fi
cannot find module providing package github.com/cybozu/neco-containers/golang/analyzer/cmd/custom-checker: working directory is not part of a module
Makefile:141: recipe for target 'custom-checker' failed
make: *** [custom-checker] Error 1
```

### DIFF
--- a/initialize/Dockerfile
+++ b/initialize/Dockerfile
@@ -3,7 +3,7 @@ FROM quay.io/cybozu/moco-mysql:8.0.20.4
 ENV GOPATH=/go
 ENV PATH=/go/bin:/usr/local/go/bin:"$PATH"
 
-ARG GO_VERSION=1.13.15
+ARG GO_VERSION=1.15.7
 USER root
 WORKDIR /work
 RUN curl -s -f -O https://dl.google.com/go/go${GO_VERSION}.linux-amd64.tar.gz \
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -5,12 +5,12 @@ on:
     branches:
       - 'master'
 env:
-  go-version: 1.13
+  go-version: 1.15
   cache-version: 1
 jobs:
   validation:
     name: Validation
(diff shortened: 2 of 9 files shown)


zoetrope (changes requested):
Don't you update go.mod and go.sum?
Especially go.sum is incompatible.

zoetrope:
I realized that go.sum is not included.

initialize/Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM quay.io/cybozu/moco-mysql:8.0.20.4
 ENV GOPATH=/go
 ENV PATH=/go/bin:/usr/local/go/bin:"$PATH"
 
-ARG GO_VERSION=1.13.15
+ARG GO_VERSION=1.15.7
```
ymmt2005, line 6: This enlarges the final image.
Why don't you split the build stage?

masa213f: This Dockerfile is used for the unit test of the `initialize` package.
The test is launched as follows (execute `go test` in the container), so it needs to be installed.

https://github.com/cybozu-go/moco/blob/v0.3.1/Makefile#L46
```
docker run -v $(PWD):/go/src/github.com/cybozu-go/moco -e GOPATH=/tmp --rm mysql-with-go:latest sh -c "CGO_ENABLED=0 go test -v ./initialize"
```

ymmt2005: I see.

zoetrope (approved):

PR #163, opened 2021-01-26
Enable the spec of unit test in clone_test.go

kfyharukz:
Signed-off-by: Kazuhito MATSUDA <kazuhito.matuda@gmail.com>

### DIFF
--- a/agent/clone_test.go
+++ b/agent/clone_test.go
@@ -462,28 +462,6 @@ func testClone() {
 			return nil
 		}, 30*time.Second).Should(Succeed())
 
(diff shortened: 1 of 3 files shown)


agent/clone_test.go
```diff
@@ -462,6 +462,28 @@ func testClone() {
 			return nil
 		}, 30*time.Second).Should(Succeed())
 
+		By("getting 500 when secret files doesn't exist")
```
masa213f, line 465: Why this test case run in this It block?
I think it is a different test case from the clone.

masa213f (approved):
Thank you!

PR #166, opened 2021-01-26
Add deletecollection for admin/edit role

masa213f:
In general, the `admin` or `editor` role is also set to `deletecollection`.
Please check the privilege of a namespace admin user by the following command.
```
kubectl -n <namespace> --as=user --as-group=<namespace admin group> --as-group=system:authenticated auth can-i --list
```

Without `deletecollection`, the team-management test in neco-apps will not succeed. :(


kfyharukz (approved):
LGTM

PR #168, opened 2021-01-28
Build kubectl-moco before e2e test

masa213f:
When `e2e/bin/kubectl-moco` already exists, the binary is not built before the e2e test.
To make it easier to do the e2e test in local, I want to fix it.


zoetrope (approved):
LGTM

Issue #172, opened 2021-01-29
Tolerance to the deletion of secret resources


masa213f:
## What

When creating MySQLCluster, the following 3 secrets are created by the moco-controller.

1. ControllerSecret (`<cluster-namespace>.<cluster-name>` in `moco-system` namespace)
  Used by the moco-controller to access mysqld instances.
2. RootPasswordSecret (`moco-root-password-<cluster-name>` in cluster's namespace)
  Mounted by mysql pods and used by agent.
3. MyCnfSecret (`moco-my-cnf-<cluster-name>` in cluster's namespace)
  Mounted by mysql pods and used for `kubectl-moco mysql` command.

Of these, 2 and 3 may be deleted by the user.
And if the secrets are deleted, some passwords will be lost and the cluster will not work.

Enable the MySQL Cluster to work even if the secrets are deleted.

## How

- Record all passwords in the ControllerSecret.
- If there the above 2 or 3 secrets are not found, regenerate them from the ControllerSecret.

## Checklist

- [x] Finish implementation of the issue
- [x] Test all functions
- [x] Have enough logs to trace activities
- [x] Notify developers of necessary actions

PR #185, opened 2021-02-06
Generate RootPasswordSecret and MyCnfSecret from ControllerSecret

d-kuro:
refs: https://github.com/cybozu-go/moco/issues/172

## Overview

* Record all passwords in the ControllerSecret.
* RootPasswordSecret and MyCnfSecret are generated from ContollerSecret.
* RootPasswordSecret and MyCnfSecret will be updated in the reconciliation roop.
  * If the User updates these secrets, they will be restored in the reconciliation loop.

### DIFF
--- a/controllers/mysqlcluster_controller.go
+++ b/controllers/mysqlcluster_controller.go
@@ -196,7 +196,13 @@ func (r *MySQLClusterReconciler) reconcileInitialize(ctx context.Context, log lo
 		return false, err
 	}
 
-	isUpdated, err = r.createSecretIfNotExist(ctx, log, cluster)
+	isUpdated, err = r.createControllerSecretIfNotExist(ctx, log, cluster)
+	isUpdatedAtLeastOnce = isUpdatedAtLeastOnce || isUpdated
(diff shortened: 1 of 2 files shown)


controllers/mysqlcluster_controller.go
```diff
@@ -456,6 +510,7 @@ func (r *MySQLClusterReconciler) createPasswordSecretForInit(ctx context.Context
 			return err
 		}
 	}
+
```
kfyharukz, line 513: The other passwords (operator, replica, donorPass) can be generated here.  I think It'll be cleaner.

kfyharukz (changes requested):

controllers/mysqlcluster_controller.go
d-kuro: Thanks!
I fixed in this commit: https://github.com/cybozu-go/moco/pull/185/commits/6425fbf9621ef0ba901ef19c1b1ffca591e2812b

kfyharukz (approved):
Looks great!

Issue #171, opened 2021-01-29
Use fluent-bit container instead of filebeat


masa213f:
## What

Use `fluent-bit`  for `err-log` and `show-log` containers in [this example]( https://github.com/cybozu-go/moco/blob/v0.4.0/docs/example_mysql_cluster.md#example).

## How

- Check that the fluent-bit is useable for our usage.

- Change the moco-controller to create StatefulSet manifest which contains err-log and slow-log containers by default.
    - Use fluent-bit image in the containers.
    - Make the containers( err-log and slow-log) customizable.

- Update documents.
   - https://github.com/cybozu-go/moco/blob/v0.4.0/docs/example_mysql_cluster.md#example
   - https://github.com/cybozu-go/moco/blob/v0.4.0/docs/design.md#how-to-manage-log-files

## Checklist

- [x] Finish implementation of the issue
- [x] Test all functions
- [x] Have enough logs to trace activities
- [x] Notify developers of necessary actions

d-kuro:
https://docs.fluentbit.io/manual/pipeline/inputs/tail

> File rotation is properly handled, including logrotate's copytruncate mode.
> Note that the Path patterns cannot match the rotated files. Otherwise, the rotated file would be read again and lead to duplicate records.

PR #215, opened 2021-03-03
Migrate from filebeat to fluent-bit

d-kuro:

This is one of the pull requests that will be split.
This pull request replaces the existing filebeat container with a fluent-bit container.
After this pull request, I will submit a change to define the default fluent-bit container and allow end users to customize the fluent-bit container.


masa213f (approved):
LGTM

PR #217, opened 2021-03-10
Inject fluent-bit container by default

d-kuro:

## About

* Automatically generates definitions for Fluent-bit containers
* Added `.spec.disableErrorLogContainer` and `.spec.disableSlowQueryLogContainer` to MySQLCluster CRD
* If the user-defined containers contain containers named `err-log` and `slow-log`, use strategic merge patch to patch them

## Docs

https://github.com/cybozu-go/moco/blob/d-kuro/inject-fluent-bit/docs/example_mysql_cluster.md#logging-containers


controllers/mysqlcluster_controller_test.go
```diff
@@ -664,6 +709,110 @@ var _ = Describe("MySQLCluster controller", func() {
 			Expect(isUpdated).Should(BeFalse())
 		})
 
+		It("should merged with user-defined fluent-bit container spec", func() {
```
kfyharukz, line 712: Please include a test to confirm the config volume can be overwritten by the user-defined configmap in this test (like the example you wrote).

kfyharukz (changes requested):

controllers/mysqlcluster_controller_test.go
kfyharukz: I withdrew the above comment. 🙇

kfyharukz (approved):
LGTM

PR #229, opened 2021-03-17
Add a default generated ConfigMap to the documentation

d-kuro:

Added fluent bit config to the documentation, which is generated by default.


kfyharukz (approved):
Looks good.

PR #240, opened 2021-05-07
Kill connections in the old primary upon a switchover

ymmt2005:
After a switchover, connections in the old primary will probably
not be usable.  To notify clients of the event earlier, MOCO should
kill the existing connections in the old primary.

The implementation does not kill connections from `localhost`
because some system processes such as `event_scheduler` have
such connections.  Also, it does not kill MOCO system users'
connections.


masa213f (approved):
LGTM

PR #261, opened 2021-06-04
Fix cloning bugs

ymmt2005:
This PR includes multiple bug fixes.

- Failed to clone large data due to a timeout.
- Failed to determine Cloning state correctly.
- Manual switchover might not have taken place immediately.

To fix the first clone bug, `moco-agent` is updated to 0.6.6.


masa213f (approved):
LGTM

PR #276, opened 2021-07-06
Fix delete BackupPolicy error

ysksuzuki:
Signed-off-by: Yusuke Suzuki <yusuke-suzuki@cybozu.co.jp>

### DIFF
--- a/api/v1beta1/backuppolicy_webhook_test.go
+++ b/api/v1beta1/backuppolicy_webhook_test.go
@@ -2,6 +2,7 @@ package v1beta1
 
 import (
 	"context"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	batchv1beta1 "k8s.io/api/batch/v1beta1"
--- a/api/v1beta1/backuppolicy_webhook_test.go
+++ b/api/v1beta1/backuppolicy_webhook_test.go
@@ -2,7 +2,6 @@ package v1beta1
 
 import (
 	"context"
-
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	batchv1beta1 "k8s.io/api/batch/v1beta1"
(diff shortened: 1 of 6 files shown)


api/v1beta1/backuppolicy_webhook_test.go
```diff
@@ -2,7 +2,6 @@ package v1beta1
 
 import (
 	"context"
-
```
ymmt2005, line 5: This line should not be removed.

backup/backup.go
```diff
@@ -418,7 +418,7 @@ func (bm *BackupManager) backupBinlog(ctx context.Context, op bkop.Operator) err
 	pw2 = nil
 
 	bw := &ByteCountWriter{}
-	key := calcKey(bm.cluster.Namespace, bm.cluster.Name, constants.BinlogFilename, lastBackup.Time.Time)
+	key := calcKey(bm.cluster.Namespace, bm.cluster.Name, constants.BinlogFilename, lastBackup.Time.Time.UTC())
```
ymmt2005, line 421: This is a standalone bug that is worth another issue or PR.
Also, I think the conversion to UTC should happen in `calcKey` because by doing so any use of `calcKey` is protected.

In fact, there is another use of `calcKey` in this same file.

backup/integration_test.go
```diff
@@ -194,7 +194,7 @@ var _ = Describe("Backup/Restore", func() {
 		Expect(bc.contents).To(HaveLen(1))
 
 		time.Sleep(1100 * time.Millisecond)
-		restorePoint := time.Now()
+		restorePoint := time.Now().UTC()
```
ymmt2005, line 197: Why do we need this conversion?

ymmt2005 (changes requested):

backup/integration_test.go
ysksuzuki: It's not necessary. Will revert.

backup/backup.go
ysksuzuki: Alright. Will create another PR for it, and convert time zone inside `calcKey`.

ymmt2005 (approved):
LGTM

PR #277, opened 2021-07-07
Fix not to be affected by the system time zone

ysksuzuki:
MySQLCluster.Status.Backup.Time is affected by the system time zone when it's retrieved.
CalcKey calculates a key for a storing backup file using the time which is affected by the system time zone, and it leads to the problem that RestoreManager can't find target backup files as expected.

Signed-off-by: Yusuke Suzuki <yusuke-suzuki@cybozu.co.jp>


ymmt2005 (approved):
LGTM

Issue #284, opened 2021-07-14
Random failure when restoring a database


ysksuzuki:
## What

A restore job can't start its restore operation because of insufficient privileges.

1. A restore job starts and waits for a mysqld ready
2. Moco-agent starts its initialization operation and creates moco-admin
3. Before moco-agent grants privileges to moco-admin, the restore job connects to the mysqld
4. The restore job can't start its restore operation, because it can't issue `SHOW MASTER STATUS` due to insufficient privileges

## How

Fix the situation described above

## Checklist

- [x] Finish implementation of the issue
- [x] Test all functions
- [x] Have enough logs to trace activities
- [x] Notify developers of necessary actions

PR #288, opened 2021-07-28
Retry restore job if the connection hasn't reflected the latest moco-admin's privileges yet

ysksuzuki:
#284

### DIFF
--- a/cmd/moco-backup/cmd/root.go
+++ b/cmd/moco-backup/cmd/root.go
@@ -65,9 +65,12 @@ var rootCmd = &cobra.Command{
 // This is called by main.main(). It only needs to happen once to the rootCmd.
 func Execute() {
 	defer func() {
-		err := recover()
-		if err != nil && errors.Is(err.(error), backup.ErrBadConnection) {
-			execute()
+		if r := recover(); r != nil {
+			if err, ok := r.(error); ok && errors.Is(err, backup.ErrBadConnection) {
+				execute()
+			} else {
+				panic(r)
+			}
 		}
 	}()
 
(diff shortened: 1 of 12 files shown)


cmd/moco-backup/cmd/root.go
```diff
@@ -63,6 +64,17 @@ var rootCmd = &cobra.Command{
 // Execute adds all child commands to the root command and sets flags appropriately.
 // This is called by main.main(). It only needs to happen once to the rootCmd.
 func Execute() {
+	defer func() {
+		err := recover()
+		if err != nil && errors.Is(err.(error), backup.ErrBadConnection) {
```
ymmt2005, line 69: if not, the error needs to be rethrown with panic.  Here's an example.
https://cs.opensource.google/go/go/+/refs/tags/go1.16.6:src/encoding/json/encode.go;l=322

ymmt2005 (changes requested):

controllers/mysqlcluster_controller.go
```diff
@@ -1092,6 +1092,11 @@ func (r *MySQLClusterReconciler) reconcileV1RestoreJob(ctx context.Context, req
 				Resources: []string{"pods"},
 				Verbs:     []string{"get"},
 			},
+			{
```
ymmt2005, line 1095: How does this relate to the issue #284?

ysksuzuki: It's nothing to do with it. I just happened to find that the restore job couldn't create events, and fixed it. Should I raise another PR?

ymmt2005: Yes, please.

ymmt2005 (approved):
LGTM

Issue #287, opened 2021-07-26
Failed to restore data due to missing PROXY privilege


shunki-fujita:
**Describe the bug**
Since moco-admin does not have PROXY privilege, if a user with PROXY privilege (for example, the root@localhost user that MySQL creates by default) exists in the database, the restore will fail.

**How**
To give full privileges to moco-admin, GRANT ALL is not enough, GRANT PROXY is also required.

https://dev.mysql.com/doc/refman/8.0/ja/grant.html
> ALL [PRIVILEGES]	Grant all privileges at specified access level except GRANT OPTION and PROXY.

User permission settings are here:
https://github.com/cybozu-go/moco-agent/blob/1ca7acd597e192f115b06426b0a82e4de850570d/server/initialize.go#L27-L29

PR #289, opened 2021-07-28
Grant PROXY privilage to AdminUser

shunki-fujita:
https://github.com/cybozu-go/moco/issues/287


ymmt2005:
The test does not pass because mysql image on DockerHub does not grant PROXY to `'root'@'%'`.

PR #309, opened 2021-10-01
Use workVolume/tmp as TMPDIR for mysqlbinlog

arosh:
When restoring from a backup, we encountered the following error.

```text
mysqlbinlog: Can't create/write to file '/tmp/tmp.HEJ6ct' (OS errno 30 - Read-only file system)
mysqlbinlog: Can't create/write to file '/tmp/tmp.fFkaQp' (OS errno 30 - Read-only file system)
mysqlbinlog: Can't create/write to file '/tmp/tmp.B2u7Xp' (OS errno 30 - Read-only file system)
mysqlbinlog: Can't create/write to file '/tmp/tmp.L82Cpt' (OS errno 30 - Read-only file system)
mysqlbinlog: Can't create/write to file '/tmp/tmp.dFxxlu' (OS errno 30 - Read-only file system)
```

mysqlbinlog requires enough space to be specified as TMPDIR. See the following.
https://dev.mysql.com/doc/refman/8.0/en/mysqlbinlog.html

### DIFF
--- a/backup/restore.go
+++ b/backup/restore.go
@@ -301,5 +301,13 @@ func (rm *RestoreManager) applyBinlog(ctx context.Context, op bkop.Operator, key
 		return fmt.Errorf("zstd exited abnormally: %w", err)
 	}
 
-	return op.LoadBinlog(ctx, binlogDir, rm.restorePoint)
+	mysqlBinlogTmpdir := filepath.Join(rm.workDir, "tmp")
+	if err := os.MkdirAll(mysqlBinlogTmpdir, 0755); err != nil {
+		return fmt.Errorf("failed to create %s: %w", mysqlBinlogTmpdir, err)
+	}
+	defer func() {
+		os.RemoveAll(mysqlBinlogTmpdir)
+	}()
+
+	return op.LoadBinlog(ctx, binlogDir, mysqlBinlogTmpdir, rm.restorePoint)
 }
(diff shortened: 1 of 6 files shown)


pkg/bkop/restore.go
```diff
@@ -93,6 +93,8 @@ func (o operator) LoadBinlog(ctx context.Context, dir string, restorePoint time.
 	binlogCmd.Stderr = os.Stderr
 	env := os.Environ()
 	env = append(env, "TZ=Etc/UTC")
+	// mysqlbinlog requires enough space to be specified as TMPDIR.
+	env = append(env, fmt.Sprintf("TMPDIR=%s", tmpDir))
```
masa213f, line 97: In this case, `+` is enough.
```suggestion
	env = append(env, "TMPDIR=" + tmpDir)
```

backup/restore.go
```diff
@@ -301,5 +301,13 @@ func (rm *RestoreManager) applyBinlog(ctx context.Context, op bkop.Operator, key
 		return fmt.Errorf("zstd exited abnormally: %w", err)
 	}
 
-	return op.LoadBinlog(ctx, binlogDir, rm.restorePoint)
+	mysqlBinlogTmpdir := filepath.Join(rm.workDir, "tmp")
```
masa213f, line 304: Compared to other variables in this function, this name is descriptive.
A simple name would be better here.

```suggestion
	tmpDir:= filepath.Join(rm.workDir, "tmp")
```

masa213f (approved):
Thank you.

PR #315, opened 2021-10-06
Add PrimaryServiceTemplate and ReplicaServiceTemplate

arosh:
## Motivation

In our Kubernetes cluster, we can automatically set desired external DNS name to the Service by annotating `external-dns.alpha.kubernetes.io/hostname`.

We want to set different annotations to the Service for primary instance and the Service for replica instances.

## How

As a patch for `serviceTemplate`, `primaryServiceTemplate` and `replicaServiceTemplate` can be specified in the manifest. They will be merged into the `serviceTemplate` applied to the corresponding service and applied in the reconciliation loop.

## Matter of concern

The merging of `serviceTemplate` and `primaryServiceTemplate` / `replicaServiceTemplate` is implemented by `DeepCopyInto`. This behavior may confuse kustomize users.

### DIFF
--- a/api/v1beta1/mysqlcluster_types.go
+++ b/api/v1beta1/mysqlcluster_types.go
@@ -36,6 +36,12 @@ type MySQLClusterSpec struct {
 	// +optional
 	ServiceTemplate *ServiceTemplate `json:"serviceTemplate,omitempty"`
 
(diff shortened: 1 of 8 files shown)


ymmt2005:
Let me close this.

arosh:
Resolved by #350

PR #339, opened 2021-11-08
Fixed a bug in decidePartSize calculation

shunki-fujita:
There are several bugs in decidePartSize:
  - If objectSize is less than **DefaultPartSize*UploadParts**, DefaultPartSize should be returned.
  https://github.com/cybozu-go/moco/blob/c9013554e92fbc97f4d29b0beb67413b16484ab9/pkg/bucket/s3.go#L139
  - Parentheses are required for shift operations.
  https://github.com/cybozu-go/moco/blob/c9013554e92fbc97f4d29b0beb67413b16484ab9/pkg/bucket/s3.go#L143

Added unittest to decidePartSize.

### DIFF
--- a/pkg/bucket/s3.go
+++ b/pkg/bucket/s3.go
@@ -13,11 +13,11 @@ import (
 )
 
 const (
-	// DefaultPartSize is the default part size used for Bucket.Put method.
-	DefaultPartSize = 128 << 20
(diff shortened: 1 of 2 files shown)


masa213f (approved):
LGTM

PR #353, opened 2021-11-26
Stop using Reconcile's context for ClusterManager

masa213f:
The context passed as a `Recincle()`'s argument is short-lived.
It is not good for the ClusterManager to depend on the short-lived context.

Regardless of the context, the ClusterManager's goroutines will be stopped here.
https://github.com/cybozu-go/moco/blob/v0.10.9/cmd/moco-controller/cmd/run.go#L86
So we need not add the another deletion function.

### DIFF
--- a/clustering/manager.go
+++ b/clustering/manager.go
@@ -23,8 +23,8 @@ import (
 //
 // This interface is meant to be used by MySQLClusterReconciler.
 type ClusterManager interface {
-	Update(context.Context, types.NamespacedName)
-	UpdateNoStart(context.Context, types.NamespacedName)
+	Update(types.NamespacedName)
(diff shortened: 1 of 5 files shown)


ymmt2005 (changes requested):
I don't think using `context.Background()` is appropriate.
The goroutine should stop when SIGINT or SIGTERM is sent to the process.

ymmt2005 (dismissed):
Understood that all goroutines will stop before returning from `subMain` because of this:
https://github.com/cybozu-go/moco/blob/v0.10.9/cmd/moco-controller/cmd/run.go#L86

ymmt2005 (approved):
LGTM

PR #356, opened 2021-12-07
backup: calculate disk usage accurately

ymmt2005:
The actual usage should be counted by the number of allocated blocks.
The new result should match with the result of `du` command.


masa213f (approved):
LGTM

PR #366, opened 2022-01-13
Do not stop gathering cluster status if one or more Pods are pending

### LINKED ISSUE #365: Status of a cluster is updated wrongly when there is a pending Pod
**Describe the bug**
While one or more Pods in a MySQL cluster StatefulSet has not been assigned an IP address,
moco-controller wrongly updates the status of the cluster to unavailable although the cluster
is in a degraded but available state.

**Environments**
- Version: v0.10.9 (the latest)

ymmt2005:
This should resolves #365


masa213f (approved):
LGTM

PR #373, opened 2022-02-20
Fixed panic when migrating to SSA.

d-kuro:
MOCO refers to the current spec of the StatefulSet when creating the SlowQueryLogContainer.

https://github.com/cybozu-go/moco/blob/abdde66ee673c7232b8c22b7396ed6665dc69ccb/controllers/mysqlcluster_controller.go#L717-L724

When I introduced server-side-apply, we used ExtractStatefulSet to get the applyconfigration, but the first time we server-side-apply, it returns an empty StatefulSetApplyConfigration because the field manager does not exist.

```
"{\"kind\":\"StatefulSet\",\"apiVersion\":\"apps/v1\",\"metadata\":{\"name\":\"moco-fuga\",\"namespace\":\"default\"}}"
```

The resulting spec was nil, which caused a panic.

https://github.com/cybozu-go/moco/blob/abdde66ee673c7232b8c22b7396ed6665dc69ccb/controllers/mysql_container.go#L164-L170

This PullRequest adds a nil check to avoid panic and to make the first time apply without field manager behave the same as when `force=true` is specified.

### DIFF
--- a/controllers/mysql_container.go
+++ b/controllers/mysql_container.go
@@ -161,14 +161,9 @@ func (r *MySQLClusterReconciler) makeV1AgentContainer(cluster *mocov1beta2.MySQL
 }
 
 func (r *MySQLClusterReconciler) makeV1SlowQueryLogContainer(sts *appsv1ac.StatefulSetApplyConfiguration, force bool) *corev1ac.ContainerApplyConfiguration {
-	isNotNil := func(sts *appsv1ac.StatefulSetApplyConfiguration) bool {
-		if sts == nil || sts.Spec == nil || sts.Spec.Template == nil || sts.Spec.Template.Spec == nil {
-			return false
-		}
-		return true
-	}
+	stsINotNil := (sts != nil && sts.Spec != nil && sts.Spec.Template != nil && sts.Spec.Template.Spec != nil)
 
-	if !force && isNotNil(sts) {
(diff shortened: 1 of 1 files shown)


controllers/mysql_container.go
```diff
@@ -161,7 +161,14 @@ func (r *MySQLClusterReconciler) makeV1AgentContainer(cluster *mocov1beta2.MySQL
 }
 
 func (r *MySQLClusterReconciler) makeV1SlowQueryLogContainer(sts *appsv1ac.StatefulSetApplyConfiguration, force bool) *corev1ac.ContainerApplyConfiguration {
-	if !force {
+	isNotNil := func(sts *appsv1ac.StatefulSetApplyConfiguration) bool {
```
ymmt2005, line 164: Why do you define a function for this?
This would be simpler.
```go
stsINotNil := !(sts == nil || sts.Spec == nil || sts.Spec.Template == nil || sts.Spec.Template.Spec == nil)
```

d-kuro: Thanks!
https://github.com/cybozu-go/moco/pull/373/commits/0ecaac56591ac5b339ab72c023fab4099de5bb75

masa213f (approved):
LGTM

zoetrope (approved):

PR #376, opened 2022-02-24
Copy values to replicaServiceTemplate in conversion webhook

d-kuro:
Copy `v1beta1.serviceTemplate` to `v1beta2.primaryServiceTempalate` and `v1beta2.replicaServiceTemplate` when converting from v1beta1 to v1beta2 with MySQLCluster conversion.

### DIFF
--- a/api/v1beta1/conversion.go
+++ b/api/v1beta1/conversion.go
@@ -5,6 +5,7 @@ import (
 	"unsafe"
 
 	"github.com/cybozu-go/moco/api/v1beta2"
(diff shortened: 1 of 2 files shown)


masa213f (changes requested):
Thank you for this PR.

I found that, in some cases, the round-trip conversion does not work.
When I create a v1beta2 MySQLCluster with `primaryServiceTemplate` only, the created resource has an unexpected `replicaServiceTemplate` field.

Could you fix it and add more tests?

masa213f (approved):
LGTM. Thank you!

Issue #306, opened 2021-08-30
Use Server-Side Apply to apply resources


zoetrope:
## What

Server-Side Apply (SSA) moves to GA in Kubernetes 1.22.
https://kubernetes.io/blog/2021/08/06/server-side-apply-ga/

So we can use SSA to apply resources.

## How

Describe how to address the issue.

## Checklist

- [ ] Finish implementation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

PR #387, opened 2022-03-09
Using SSA to create Secret

d-kuro:
refs: https://github.com/cybozu-go/moco/issues/306

Create Secret using server-side-apply


masa213f (approved):
LGTM

PR #388, opened 2022-03-09
Validate that container name is non-nil in validation webhook

d-kuro:
refs: #306 

Validate that the container name is non-nil with the validation webhook.
This is a follow-up to the validation that disappeared from the CRD spec when SSA was introduced.


masa213f (approved):
LGTM

Issue #235, opened 2021-04-26
Guaranteed QoS


ymmt2005:
## What

Currently, MOCO users cannot make MySQL pods to have [Guaranteed QoS](https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/#create-a-pod-that-gets-assigned-a-qos-class-of-guaranteed) because MOCO does not limit resources to some containers.

Improve MOCO to allow users to make Guaranteed MySQL Pods.

## How

Describe how to address the issue.

## Checklist

- [ ] Finish implementation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

arosh:
According to our experience of running MOCO for the past half year in our development environment, the max memory usages by containers except for mysqld are as follows. The memory usages are observed by `container_memory_working_set_bytes` from cAdvisor.

| container | max memory usage |
| ---------- | ---------------------- |
| agent (moco-agent) | 60 MiB |
| slow-log (fluent-bit) | 9 MiB |
| mysqld_exporter | 38 MiB |

PR #394, opened 2022-03-22
Add design doc for allow users to customize containers

d-kuro:
refs: https://github.com/cybozu-go/moco/issues/235

### DIFF
--- a/docs/designdoc/allow_customize_containers.md
+++ b/docs/designdoc/allow_customize_containers.md
@@ -1,0 +1,104 @@
+# Allow users to customize containers
+
+## Context
+
+Currently, MOCO has containers that are automatically added by the system in addition to containers added by the user.
+(e.g. `agent`, `moco-init` etc...)
+Containers automatically added by the system do not allow user customization.
+
+There are cases where users want to customize container resources, as in Issue [cybozu-go/moco#235](https://github.com/cybozu-go/moco/issues/235).
+This design document examines how users can customize containers that are automatically added by the system.
+
+## Goals
+
+* No breaking changes
+* Allows users to customize containers that are automatically added by the system
+* Extendable functionality for future customization beyond container resources
+* Users can only customize authorized fields
+  * Do not allow customization of fields that would make operation untenable (e.g. `command`)
+
+## Non-goals
+
+* No customization provided at the time of implementation other than requested container resources
+* No enhancement will be added to the `moco.cybozu.com/v1beta1` API.
+
+## ActualDesign
+
+Add `.spec.podTemplate.overwriteContainers` field to MySQLCluster.
+
+```yaml
+spec:
+  podTemplate:
+    spec:
+      containers:
+      - name: mysqld
+        image: quay.io/cybozu/mysql:8.0.28
+    overwriteContainers:
+    - name: agent
+      resources:
+        requests:
+          cpu: 50m
+```
+
+Pseudo-code:
+
+```go
+type PodTemplateSpec struct {
+	ObjectMeta
+	Spec                PodSpecApplyConfiguration
+	OverwriteContainers []OverwriteContainer // added
+}
+
+type OverwriteContainer struct {
+	Name      string // required
+	Resources *corev1ac.ResourceRequirementsApplyConfiguration
+}
+```
+
+`overwriteContainers` is a container definition with only name and customizable fields.
+moco-controller refers to `overwriteContainers` when creating containers and overwrites fields if the container name matches.
+
+`overwriteContainer` does not distinguish between initContainer and container, and the names of required fields are validated with the validation webhook.
+No merge logic is provided for container resources to avoid implicit value setting situations.
+
+Pros:
+
+* Expandable
+* Users can learn about customizable fields from MySQLCluster specs
+
+Cons:
+
+* If you want different customizable fields for different containers, you cannot express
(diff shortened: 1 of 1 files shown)


docs/designdoc/allow_customize_containers.md
```diff
@@ -0,0 +1,104 @@
...
+* Expandable
+* Users can learn about customizable fields from MySQLCluster specs
+
+Cons:
+
+* If you want different customizable fields for different containers, you cannot express
+* Users need to know the list of containers to be added by the system
+  * Follow it in the documentation
+
+## AlternativesConsidered
```
ymmt2005, line 75: I prefer this alternative design.  It is at least explicit.
We can reduce potential user stress by setting the default resource limits.

docs/designdoc/allow_customize_containers.md
```diff
@@ -0,0 +1,104 @@
...
+type OverwriteContainer struct {
+	Name      string // required
+	Resources *corev1ac.ResourceRequirementsApplyConfiguration
+}
+```
+
+`overwriteContainers` is a container definition with only name and customizable fields.
+moco-controller refers to `overwriteContainers` when creating containers and overwrites fields if the container name matches.
+
+`overwriteContainer` does not distinguish between initContainer and container, and the names of required fields are validated with the validation webhook.
```
ymmt2005, line 61:
```suggestion
`overwriteContainers` does not distinguish between initContainer and container, and the names of required fields are validated with the validation webhook.
```

ymmt2005 (changes requested):

docs/designdoc/allow_customize_containers.md
masa213f: @ymmt2005 
I discussed this design with @zoetrope. And we prefer the first one.
We think it is easy to approach because the structure is similar to the `corev1.Container`.

Could we choose your proposed alternative design, which limits the `name` filed with the `Enum` marker?

https://book.kubebuilder.io/reference/markers/crd-validation.html
```
// +kubebuilder:validation:Enum=<[]any>
    specifies that this (scalar) field is restricted to the *exact* values specified here.
```

ymmt2005: @masa213f Not bad :-)

masa213f: @ymmt2005 
Thanks for checking!

@d-kuro 
Could you update the first design to limit with the enum marker?

docs/designdoc/allow_customize_containers.md
```diff
@@ -0,0 +1,105 @@
...
+type PodTemplateSpec struct {
+	ObjectMeta
+	Spec                PodSpecApplyConfiguration
+	OverwriteContainers []OverwriteContainer // added
+}
+
+type OverwriteContainer struct {
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum=agent;moco-init;slow-log;mysqld-exporter
+	Name      string
```
ymmt2005, line 55: Actually, this should be defined as a dedicated type like:

```go
// +kubebuilder:validation:Enum=agent;moco-init;slow-log;mysqld-exporter
type ContainerName string

const (
    AgentName ContainerName = "agent"
    ...
)
```

Also, this definition should replace the previous definition in `pkg/constants`.
https://github.com/cybozu-go/moco/blob/main/pkg/constants/container.go#L3

Please replace the references of the previous definition with this one.

d-kuro: @ymmt2005 
Thanks for the comments!
Since the container names defined in pkg/constants are not all overwritable containers (I'm thinking that the mysqld container is user-specifiable, so there is no need to define it as an overwritable container) I was thinking of adding the following definition to the v1beta2 package, what do you think?

```go
type PodTemplateSpec struct {
	// Standard object's metadata.  The name in this metadata is ignored.
	// +optional
	ObjectMeta `json:"metadata,omitempty"`

	// Specification of the desired behavior of the pod.
	// The name of the MySQL server container in this spec must be `mysqld`.
	Spec PodSpecApplyConfiguration `json:"spec"`

	// OverwriteContainers overwrites the container definitions provided by default by the system.
	// +optional
	OverwriteContainers []OverwriteContainer `json:"overwriteContainers,omitempty"`
}

// +kubebuilder:validation:Enum=agent;moco-init;slow-log;mysqld-exporter
type OverwriteableContainerName string

const (
	AgentContainerName             OverwriteableContainerName = constants.AgentContainerName
	InitContainerName              OverwriteableContainerName = constants.InitContainerName
	SlowQueryLogAgentContainerName OverwriteableContainerName = constants.SlowQueryLogAgentContainerName
	ExporterContainerName          OverwriteableContainerName = constants.ExporterContainerName
)

type OverwriteContainer struct {
	// +kubebuilder:validation:Required
	Name OverwriteableContainerName `json:"name"`

	// +optional
	Resources *ResourceRequirementsApplyConfiguration `json:"resources,omitempty"`
}
```

ymmt2005: Looks good!

d-kuro: Updated: https://github.com/cybozu-go/moco/pull/394/commits/befa8c856c659e77952180891251b55153ee6b6a

ymmt2005 (approved):
LGTM

masa213f (approved):

PR #416, opened 2022-06-06
Use Docker action

inductor:
This PR improves Docker build cache using Docker official build action instead of using shell script.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -6,27 +6,74 @@ on:
 env:
   go-version: 1.17
 jobs:
-  image:
+  build-moco:
     name: Push Container Image
     runs-on: ubuntu-20.04
     steps:
     - uses: actions/checkout@v3
-    - name: Build images
-      run: |
-        docker build -t moco:dev --target controller .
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -6,27 +6,74 @@ on:
...
+        images: ghcr.io/cybozu-go/moco
+        # ref. https://github.com/docker/metadata-action#typesemver
+        tags: |
+          type=semver,pattern={{version}}
+    - name: Build (and push if on main)
+      id: docker_build
+      uses: docker/build-push-action@v2
+      with:
+        builder: ${{ steps.buildx.outputs.name }}
+        push: true
```
inductor, line 36: You can also test docker build on each commit if you set this as false on other action triggers.

morimoto-cybozu (approved):
Thank you for your contribution!
As we are migrating to Docker's GitHub Actions, your commit is very helpful for us.
We welcome your feedback and further commits:-)

umezawatakeshi (approved):
Seems good. Thank you!

Issue #265, opened 2021-06-21
Document the procedure to increase volume size


ymmt2005:
## What

Kubernetes currently does not allow users to increase the storage sizes in `spec.volumeClaimTemplates`
of an already created StatefulSet.  There is an ongoing [KEP](https://github.com/kubernetes/enhancements/pull/1848), though.

Since MOCO creates a StatefulSet for each MySQL cluster, users cannot increase the storage size declared in
`spec.volumeClaimTemplates[].spec.resources.requests["storage"]` of MySQLCluster.

To increase the size, users need to follow the instructions shown below:

1. Resize all PVCs manually
2. Edit MySQLCluster and change the requested volume size
3. Delete StatefulSet w/o deleting Pods
    - `kubectl delete sts moco-xxx --cascade=orphan`

Afterward, `moco-controller` will successfully recreate StatefulSet with the new size.

## How

Describe how to address the issue.

## Checklist

- [ ] Finish implementation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

ClaudiuLSH:
Can't this be automated? Running manual operations on a production deployment should be limited as much as possible maybe?

[Here is how another project handled this](https://github.com/zalando/postgres-operator/blob/a6d4f23f90fe12822a19e758ff400233c9212e00/pkg/cluster/volumes.go#L46).

ymmt2005:
@ClaudiuLSH Yes, this can be automated.

Such a feature enhancement is preferable to documentation like this.

ymmt2005:
@d-kuro can we close this now?

d-kuro:
@ymmt2005 thanks!, I have completed the task and will close

PR #421, opened 2022-07-06
Add statefulset invalid error metrics

d-kuro:
refs: #265

Added metrics to show re-creation counts when StatefulSet re-creation is supported.

This Pull Request adds metrics for the number of times an update of the immutable field in the StatefulSet has resulted in an error.
This allows users to detect unrecoverable StatefulSet update errors.

### DIFF
--- a/controllers/mysqlcluster_controller.go
+++ b/controllers/mysqlcluster_controller.go
@@ -968,6 +968,9 @@ func (r *MySQLClusterReconciler) reconcileV1StatefulSet(ctx context.Context, req
 		if needRecreate {
 			metrics.StatefulSetRecreateErrorTotal.WithLabelValues(cluster.Name, cluster.Namespace).Inc()
 		}
+		if apierrors.IsInvalid(err) {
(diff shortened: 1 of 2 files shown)


controllers/mysqlcluster_controller.go
```diff
@@ -968,6 +968,9 @@ func (r *MySQLClusterReconciler) reconcileV1StatefulSet(ctx context.Context, req
 		if needRecreate {
 			metrics.StatefulSetRecreateErrorTotal.WithLabelValues(cluster.Name, cluster.Namespace).Inc()
 		}
+		if apierrors.IsInvalid(err) {
```
d-kuro, line 971: I was going to write a test, but couldn't figure out how to cause this with the current MySQLCluster spec.
Is there a better way?

masa213f: @d-kuro 
Sorry, I misunderstood the specification of the MySQLCluster.
Certainly, the invalid update will not happen...

May I close this PR?

PR #429, opened 2022-07-20
Support kubernetes v1.24

d-kuro:
Support Kubernetes v1.24.

* update deps
* update tools
* update Go version to v1.18


api/v1beta1/conversion_test.go
```diff
@@ -18,75 +20,83 @@ func TestCompatibility(t *testing.T) {
...
+		}
+		if err := scheme.Convert(&tmp1, &v1beta2Cluster, nil); err != nil {
+			t.Fatal(err)
+		}
+		if err := scheme.Convert(&v1beta2Cluster, &tmp2, nil); err != nil {
+			t.Fatal(err)
+		}
+		if err := scheme.Convert(&tmp2, &v1beta1Cluster2, nil); err != nil {
+			t.Fatal(err)
+		}
```
masa213f, line 46: Could you tell me why we converted it four times? Sorry, I forgot :(

d-kuro: Certainly no need for 4 conversions...

When I implemented conversion in the past, I tried to prepare internal objects in addition to v1beta1 and v1beta2.
In the controller logic, using an internal objcet has the advantage of not having to change the api version used by the controller when the api version increases in the future.

I prepare an internal object, you will need to do the conversion 4 times:

* v1beta1 -> internal -> v1beta2 -> internal -> v1beta1

However, the implementation became so complex that we decided not to adopt internal objects.
I had forgotten to make the fix to the test at that time.

d-kuro: I will modify it to the proper conversion logic.
Thanks for review 🙏

d-kuro: fixed: https://github.com/cybozu-go/moco/pull/429/commits/27856251cf9b6b066f98a58853b213e0d9e68767

masa213f (approved):
LGTM

Issue #407, opened 2022-04-25
Is it not acceptable to disable maxDelaySeconds?


arosh:
**Describe the bug**

Even if maxDelaySeconds: 0 is configured to disable the functionality to mark mysqld with delayed transaction application as non-ready, the default value maxDelayThreshold: 60 is still configured for the readiness probe.

**Environments**
- Version: 0.11.1 (chart-v0.2.1)
- OS: Flarcar Container Linux (stable)

**To Reproduce**

1. Deploy a MySQLCluster with maxDelaySeconds: 0
2. Execute huge transactions where the replication lag exceeds 60 seconds
3. Browsing the moco-agent logs we find maxDelayThreshold: 60 and msg: "the instance delays from the primary"

**Expected behavior**

[The description in the documentation](https://cybozu-go.github.io/moco/crd_mysqlcluster_v1beta2.html?highlight=maxDelaySeconds#sub-resources), "if set, configures the readiness probe" makes us expect the feature to be disabled if this value is not set.

The following lines make it expected that maxDelaySeconds: 0 will disable the feature.

* https://github.com/cybozu-go/moco-agent/blob/348f568b0c88ad5330fc359b84968fd281402d02/server/mysqld_health.go#L79-L83
* https://github.com/cybozu-go/moco/blob/e13dde591cf5a23287c1c44c1672b7656aa27610/controllers/mysql_container.go#L114

**Additional context**

The reason why 60 seconds is set to the readiness probe even though maxDelaySeconds: 0 is due to the following line.

https://github.com/cybozu-go/moco-agent/blob/348f568b0c88ad5330fc359b84968fd281402d02/cmd/moco-agent/cmd/root.go#L236

morimoto-cybozu:
The phrase "MaxDelaySeconds, if set, configures the readiness probe" is inappropriate.
With the current implementation, the threshold is configured to 60s if `MaxDelaySeconds` is not set.

I'll fix the description to "MaxDelaySeconds configures the readiness probe" first.
The default will still be 60s.

The effect of explicit `maxDelaySeconds: 0` would need a careful discussion.
"0" seems suitable for disabling the check, while "max delay == 0" in a literal interpretation means the most strict check.
I'll adopt the former.

PR #434, opened 2022-07-26
Disable the delay check if .spec.maxDelaySeconds == 0 in MySQLCluster

morimoto-cybozu:
This commit clarifies and changes the behavior of the readiness probe
for the case when .spec.maxDelaySeconds == 0 in MySQLCluster.
Before this, setting the field to 0 is considered as specifying
the default value, i.e. 60s.
Now setting the field to 0 disables the delay check in the probe.

Fixes: #407


umezawatakeshi (approved):
LGTM

Note: [moco-agent](https://github.com/cybozu-go/moco-agent)'s document doesn't seem to say that `--max-delay=0` disables delay check.

Issue #445, opened 2022-08-29
Setting max_connect_errors is meaningless


shunki-fujita:
## What
MOCO enables `skip_name_resolve` by default.
https://github.com/cybozu-go/moco/blob/d4afdf307ad0355215e46111e23d5660e4ae15d7/pkg/mycnf/generator.go#L104

The `max_connect_errors` option will have no effect because its behavior depends on the host cache, which is disabled by skip_name_resolve.
https://github.com/cybozu-go/moco/blob/d4afdf307ad0355215e46111e23d5660e4ae15d7/pkg/mycnf/generator.go#L31

## How
Remove `max_connect_errors` setting as it is meaningless.

## Checklist

- [ ] Finish implementation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

masa213f:
Thank you. merged.

PR #446, opened 2022-08-29
issue-445: Remove max_connect_errors setting as it is meaningless

shunki-fujita:
https://github.com/cybozu-go/moco/issues/445


ymmt2005 (approved):
LGTM

PR #447, opened 2022-08-30
Remove batch/v1beta1 api

d-kuro:
refs: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-25

> The batch/v1beta1 API version of CronJob will no longer be served in v1.25.

Migrate cronjob used by moco from `batch/v1beta1` to `batch/v1`.


masa213f (approved):
LGTM

PR #465, opened 2022-09-27
Use reusing workflow

d-kuro:
There are many versions of MySQL supported by MOCO.

> MySQL: 8.0.18, 8.0.25, 8.0.26, 8.0.27, 8.0.28, 8.0.30

Since the matrix test will continue to grow if this is not done, the following policy is established.

* Determine the MySQL version of the LTS
  * This is the version that Cybozu uses internally
* In addition to the LTS version, CI will run the latest and one of the previous versions
* Matrix tests for all supported MySQL versions are run on a weekly basis

### DIFF
--- a/.github/workflows/ci-e2e.yaml
+++ b/.github/workflows/ci-e2e.yaml
@@ -22,7 +22,7 @@ jobs:
     name: Integration tests with MySQL
     strategy:
       matrix:
-        mysql-version: ["8.0.18", "8.0.28", "8.0.30"]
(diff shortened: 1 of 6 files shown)


masa213f (changes requested):
Thank you for the good improvements!

Please make a few changes.

- We use MySQL 8.0.28 internally. So MySQL 8.0.18 no longer needs to test in each PR.
- Basically, we continue to maintain all supported MySQL. I'm afraid that the word `LTS` is not suitable. Could you replace it with another word in some comments?

masa213f (approved):
LGTM!

Issue #463, opened 2022-09-27
moco-controller keeps restarting


uqix:
**Describe the bug**
moco-controller keeps restarting.

**Environments**
- Version: helm 0.3.0
- OS: k8s v1.20.12, centos 7.8

**Additional context**
* moco-controllers run on nodes without internet connection

Error logs:

```
{"level":"error","ts":1664248990.236583,"logger":"controller-runtime.source","msg":"if kind is a CRD, it should be installed before calling Start","kind":"PodDisruptionBudget.policy","error":"no matches for kind \"PodDisruptionBudget\" in version \"policy/v1\"","stacktrace":"sigs.k8s.io/controller-runtime/pkg/source.(*Kind).Start.func1.1\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/source/source.go:139\nk8s.io/apimachinery/pkg/util/wait.runConditionWithCrashProtectionWithContext\n\t/go/pkg/mod/k8s.io/apimachinery@v0.24.3/pkg/util/wait/wait.go:233\nk8s.io/apimachinery/pkg/util/wait.WaitForWithContext\n\t/go/pkg/mod/k8s.io/apimachinery@v0.24.3/pkg/util/wait/wait.go:660\nk8s.io/apimachinery/pkg/util/wait.poll\n\t/go/pkg/mod/k8s.io/apimachinery@v0.24.3/pkg/util/wait/wait.go:594\nk8s.io/apimachinery/pkg/util/wait.PollImmediateUntilWithContext\n\t/go/pkg/mod/k8s.io/apimachinery@v0.24.3/pkg/util/wait/wait.go:545\nsigs.k8s.io/controller-runtime/pkg/source.(*Kind).Start.func1\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/source/source.go:132"}
...
Error: failed to wait for mysqlcluster caches to sync: timed out waiting for cache to be synced
```

Full logs:

see attachment: [moco-controller-7bbd5bb6f9-8246z.log](https://github.com/cybozu-go/moco/files/9651260/moco-controller-7bbd5bb6f9-8246z.log).

uqix:
Downgrading to helm 0.2.3 works

ymmt2005:
@d-kuro could you take a look?

d-kuro:
@uqix 

Chart version v0.3.0 uses MOCO v0.13.0 by default.[^1]

[...]

PR #468, opened 2022-10-11
Specify minimum Kubernetes version

d-kuro:
refs: https://github.com/cybozu-go/moco/issues/463

Current Helm chart does not work with Kubernetes v1.21 or lower by default.
Specify the minimum Kubernetes version.

e.g.

```console
$ helm install moco ./charts/moco/
Error: INSTALLATION FAILED: chart requires kubeVersion: >= 1.22.0 which is incompatible with Kubernetes v1.21.12
```


masa213f (approved):
LGTM

Issue #413, opened 2022-05-16
Avoid running backup/restore pods on the same host in parallel


ymmt2005:
## What

Taking a backup of a MySQL instance is often resource-consuming.
It would be great if users of MOCO can avoid running backup / restore Pods running on the same host in parallel.

## How

Add `affinity` stanza to jobConfig kind and set the affinity to the Pod template of Jobs.

## Checklist

- [ ] Finish implementation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

PR #471, opened 2022-10-19
Add affinity for BackupPolicySpec

d-kuro:
refs: https://github.com/cybozu-go/moco/issues/413


ymmt2005:
Could you also create an example manifest in `examples` directory?
https://github.com/cybozu-go/moco/tree/main/examples

d-kuro:
@ymmt2005 Thanks review, sorry for late the reply.
Question, shouldn't it add anti-affinity by default?

https://github.com/cybozu-go/moco/pull/471/commits/57e56b4a99b087ea3e36832d2472be22fd4714d3

ymmt2005:
@d-kuro 
Good point. I think it's a good idea to set an anti-affinity by default as long as its policy is `preferredDuringSchedulingIgnoredDuringExecution`.

@masa213f what do you think?

masa213f:
@ymmt2005 @d-kuro 
Sorry for the late reply.
That's good idea to set the default anti-affinity of preferredDuringSchedulingIgnoredDuringExecution.

d-kuro:
@ymmt2005 @masa213f 
Sorry for the delay in responding.
I added default anti-affinity with `preferredDuringSchedulingIgnoredDuringExecution` and update docs.
https://github.com/cybozu-go/moco/pull/471/commits/bf10dbf8fef64b53c99b2ca02c90b4bd83043eb2

pkg/constants/meta.go
```diff
@@ -10,6 +10,8 @@ const (
 	LabelAppCreatedBy = "app.kubernetes.io/created-by"
 	AppCreator        = "moco"
 
+	LabelHostName = "kubernetes.io/hostname"
```
ymmt2005, line 13: I think we can use `corev1.LabelHostname`.
https://github.com/kubernetes/api/blob/3a2d6b5bb7f84daa2125131865eb81b151448d5b/core/v1/well_known_labels.go#L20

ymmt2005 (changes requested):

masa213f (dismissed):
LGTM. Please fix https://github.com/cybozu-go/moco/pull/471#discussion_r1048036669

pkg/constants/meta.go
d-kuro: Thanks! I fixed: https://github.com/cybozu-go/moco/pull/471/commits/bef9388871be165040b78b4d277169df335aee70

e2e/mycluster.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: moco.cybozu.com/v1beta1
```
masa213f, line 1: Do we need this file?

masa213f (changes requested):

e2e/mycluster.yaml
d-kuro: I certainly may not need it.
I will remove it.

d-kuro: Sorry, I committed another unnecessary file.
I remove it.

ymmt2005 (dismissed):

e2e/mycluster.yaml
d-kuro: I deleted the file and squashed the commit.
https://github.com/cybozu-go/moco/pull/471/commits/4548faf31bb4d5341048be6d40add60478a264ce

masa213f (approved):
LGTM

Issue #361, opened 2021-12-24
Terminate MySQL container faster when probe fails


morimoto-cybozu:
## What

MOCO configures `.spec.template.spec.terminationGracePeriodSeconds` of a MySQL StatefulSet to `300` by default.
This is good for ordinary termination, but too long for abnormal termination caused by probe failure.

The per-probe `terminationGracePeriodSeconds` introduced in K8s 1.21 would be useful for abnormal termination.

Note that this property is available only if the Feature Gate `ProbeTerminationGracePeriod` is enabled.
If the Feature Gate is disabled, the property is accepted in Create/Update but not returned in Get requests.
To avoid infinite reconciliation retries, the MOCO controller should provide a configuration method to enable/disable per-probe `terminationGracePeriodSeconds`.

The `ProbeTerminationGracePeriod` feature gate became beta in Kubernetes 1.22.
After we drop supports for Kubernetes 1.21 and prior versions, we may use this feature unconditionally.

## How

Describe how to address the issue.

## Checklist

- [ ] Finish implementation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

d-kuro:
@morimoto-cybozu 
Now available in moco v0.13.0.
https://github.com/cybozu-go/moco/blob/v0.13.0/config/crd/bases/moco.cybozu.com_mysqlclusters.yaml#L916

morimoto-cybozu:
@d-kuro 
That line is a part of `.spec.podTemplate` of the `MySQLCluster` CRD, so it means that the users of MOCO can configure the per-probe `terminationGracePeriodSeconds` manually by themselves.
It would be good if the MOCO controller configures the per-probe `terminationGracePeriodSeconds` as it does for the per-pod one.
https://github.com/cybozu-go/moco/blob/v0.13.0/controllers/mysqlcluster_controller.go#L819

PR #472, opened 2022-10-26
Uses a probe defined by the user

d-kuro:
refs: https://github.com/cybozu-go/moco/issues/361

Allows the user to configure mysql container probe.


masa213f (approved):
LGTM

Issue #419, opened 2022-06-27
(Short time) network disconnection on primary instance could mislead MOCO into believing that errant transactions have occurred on replica instances


arosh:
**Describe the bug**

In our environment, we experienced a brief network disconnection (for reasons unrelated to MOCO) and received the following errors.

> **primary**
> 2022-06-24T04:15:53.924593Z 0 [ERROR] [MY-013129] [Server] A message intended for a client cannot be sent there as no client-session is attached. Therefore, we're sending the information to the error-log instead: MY-001158 - Got an error reading communication packets
> 2022-06-24T04:15:53.925788Z 16406 [Note] [MY-011171] [Server] Stop semi-sync binlog_dump to slave (server_id: 805432148).
2022-06-24 13:15:53

> **replica**
> 2022-06-24T04:15:53.894865Z 16087 [Note] [MY-011026] [Repl] Slave I/O thread killed while reading event for channel ''.
> 2022-06-24T04:15:53.894958Z 16087 [Note] [MY-010570] [Repl] Slave I/O thread exiting for channel '', read up to log 'binlog.000121', position 807839346

No failover occurred because the network disconnection was immediately resolved. However, MOCO detected an errant transaction in the replica (moco_cluster_errant_replicas{} = 2) and declined to execute a START SLAVE. Investigation using SHOW BINLOG EVENTS did not find any errant transactions.

**Environments**
- Version: v0.12.1 (chart-v0.2.3)
- OS: Flatcar Container Linux (stable)

**To Reproduce**
1. Create MySQLCluster
2. Use benchmarking tools to ensure that transactions are constantly being committed
3. Disconnect the network connection between the primary instance and the replica instances
4. See moco_cluster_errant_replicas

**Expected behavior**
In evaluating an errant transaction, the primary instance is treated as having the GTID of a transaction that has been committed to binlog but not committed to InnoDB.

**Additional context**
This behavior can be occured in loss-less semisynchronous replication. In this configuration, the primary instance commits the transaction to InnoDB only after confirming that the replica has received the transaction's binlog.

Consider the case where a network disconnection occurs while a replica is sending packets acknowledging receipt of transaction binlog. The replica SQL thread can apply the received binlog to its own InnoDB, whereas the primary cannot commit the transaction to InnoDB due to the constraint of loss-less semisynchronous replication.

ymmt2005:
A possible workaround would be to `START SLAVE` anyway when the replica seems to have an extra transaction.

[...]

PR #491, opened 2022-12-15
Ignore primary's event when detecting errant replicas

masa213f:
refs: https://github.com/cybozu-go/moco/issues/419

When the primary MySQL load is high, in the rare case, it will delay the execution of transactions in the primary, and the `gtid_executed` of replicas precedes the primary.
In this case, the moco-controller recognizes preceding replicas as errant replicas and purges them from a MySQLCluster.
We have faced this situation twice, and in both cases, the moco-controller purged all replicas, and the MySQLCluster became unavailable.

This PR fixes the detection logic for errant transactions to prevent cluster collapse as described above.
We allow replicas to precede the primary only one event.

### DIFF
--- a/clustering/status.go
+++ b/clustering/status.go
@@ -261,7 +261,7 @@ func (p *managerProcess) GatherStatus(ctx context.Context) (*StatusSet, error) {
 }
 
 // containErrantTransactions check whether a GTID set contains errant transactions.
-// When the primary load is high, in the rare case, gtid_executed of replicas precedes the primary. (Probably only one event)
+// When the primary load is high, in the rare case, gtid_executed of replicas precedes the primary.
 // Assuming such a situation, this function ignores primary's event.
 func containErrantTransactions(primaryUUID, gtidSet string) bool {
 	if primaryUUID == "" {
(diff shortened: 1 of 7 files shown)


clustering/status.go
```diff
@@ -258,6 +260,45 @@ func (p *managerProcess) GatherStatus(ctx context.Context) (*StatusSet, error) {
...
+
+	uuidSets := strings.Split(gtidSet, ",")
+	if len(uuidSets) > 1 {
+		return true
+	}
+	split := strings.Split(uuidSets[0], ":")
+	if split[0] != primaryUUID {
+		return true
+	}
+	// If two or more events exist, they are considered errant transactions.
```
ymmt2005, line 297: I'm a bit concerned about this because MySQL has a feature called "group commit".
https://dev.mysql.com/worklog/task/?id=5223

With the group commit, multiple transactions can be committed at once.

masa213f: Thank you!
As you pointed out, there is a possibility that multiple events will be executed at once by "group commit".
So I will allow multiple events as well.

72e68ca

ymmt2005 (approved):
LGTM

ymmt2005 (approved):

Issue #427, opened 2022-07-15
Backup to Google Cloud Storage


filiprafaj:
Hello, I am trying and failing to store backups to Google Cloud Storage.

I have set up [Workload Identity](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity) to give the k8s service account the permissions to access the bucket.

I have also tried defining the _AWS_ACCESS_KEY_ID_ and _AWS_SECRET_ACCESS_KEY_ variables.

Still I get
`Error: failed to take a full dump: failed to put dump.tar: operation error S3: PutObject, https response error StatusCode: 403, RequestID: , HostID: , api error SignatureDoesNotMatch: The request signature we calculated does not match the signature you provided. Check your Google secret key and signing method.`

Is there anybody who made backups to GCS working? 

Thank you!

ymmt2005:
Backup to GCS is not supported now.
It can be added by implementing Bucket interface for GCS in this package.
https://github.com/cybozu-go/moco/tree/main/pkg/bucket

We are welcome to a pull request for adding GCS support.

d-kuro:
@filiprafaj 
GCS supports the S3 compatibility API. (Sorry, I have not verified this.)
Could you please refer to the documentation and try again?

refs: 
* https://cloud.google.com/storage/docs/interoperability#xml_api
* https://vamsiramakrishnan.medium.com/a-study-on-using-google-cloud-storage-with-the-s3-compatibility-api-324d31b8dfeb

If it still doesn't work, it would be helpful if you could report it again, including the definition of BackupPolicy 🙏 
https://cybozu-go.github.io/moco/usage.html#backuppolicy

filiprafaj:
Hi @d-kuro , I have tried now with **interoperability credentials** and I am getting:

`
Error: failed to take a full dump: failed to put dump.tar: operation error S3: PutObject, https response error StatusCode: 403, RequestID: , HostID: , api error SignatureDoesNotMatch: The request signature we calculated does not match the signature you provided. Check your Google secret key and signing method. failed to take a full dump: failed to put dump.tar: operation error S3: PutObject, https response error StatusCode: 403, RequestID: , HostID: , api error SignatureDoesNotMatch: The request signature we calculated does not match the signature you provided. Check your Google secret key and signing method. 
`

the backup file looks like this:

```
apiVersion: moco.cybozu.com/v1beta2
kind: BackupPolicy
metadata:
  namespace: default
  name: daily
spec:
[...]

PR #493, opened 2022-12-21
Added proposal to support non-S3 compatible object storage

d-kuro:
refs: https://github.com/cybozu-go/moco/issues/427

Propose API enhancements to allow GCS and other non-S3 compatible object storage to be used from MOCO

### DIFF
--- a/docs/designdoc/object_storage_type.md
+++ b/docs/designdoc/object_storage_type.md
@@ -1,0 +1,82 @@
+# Make object storage selectable
+
+## Context
+
+The only object storage interface currently supported by MOCO for storing backups is Amazon S3.
+Amazon S3 interfaces are available for many object storage.
+As a result, MOCO is currently able to support many object storage.
+
+However, there are cases that are not supported.
+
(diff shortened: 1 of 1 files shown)


docs/designdoc/object-storage-type.md
```diff
@@ -0,0 +1,55 @@
...
+apiVersion: moco.cybozu.com/v1beta2
+kind: BackupPolicy
+metadata:
+  namespace: backup
+  name: daily
+spec:
+  jobConfig:
+...
+    bucketConfig:
+      backendType: s3 # added
```
ymmt2005, line 50: Will `s3` be the default?

d-kuro: Yes.
I think it is appropriate to default to the S3 API as it has become the de facto standard interface for almost all storage providers.

I'm considering setting the string `s3` if empty in the mutating webhook.

ymmt2005: Good. Note also that existing BackupPolicy might have empty `backendType`,
so MOCO should treat empty as S3.

masa213f (approved):
LGTM

PR #500, opened 2023-01-27
Add pprof server to moco-controller

masa213f:
Add pprof server to moco-controller.
The pprof server will start only when `--pprof-addr` option is set.

> **Note**
> We can remove the `pprof.Handler` added in this PR after merging the following PR into the controller runtime.
> https://github.com/kubernetes-sigs/controller-runtime/pull/1943

Usage: View the trace data
1. Deploy the moco-controller with `--pprof-addr=:9090` option.
2. Get trace data and run the trace viewer.
    ```bash
    # Get trace data
    $ kubectl port-forward -n moco-system POD_NAME  9090:9090 &
    $ curl -s http://localhost:9090/debug/pprof/trace?seconds=10 > trace.out

    # Run trace viewer
    $ go tool trace -http="localhost:9091" trace.out 
    ```
3. Open `http://localhost:9091` with a Web brower.

### DIFF
--- a/cmd/moco-controller/cmd/run.go
+++ b/cmd/moco-controller/cmd/run.go
@@ -12,6 +12,7 @@ import (
 	"github.com/cybozu-go/moco/pkg/cert"
 	"github.com/cybozu-go/moco/pkg/dbop"
 	"github.com/cybozu-go/moco/pkg/metrics"
+	"github.com/cybozu-go/moco/pkg/pprof"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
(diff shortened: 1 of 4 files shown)


pkg/pprof/pprof.go
```diff
@@ -0,0 +1,65 @@
...
+
+func (h *Handler) Start(ctx context.Context) error {
+	mux := http.NewServeMux()
+	mux.HandleFunc("/debug/pprof/", pprof.Index)
+	mux.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
+	mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
+	mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
+	mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
+
+	server := &http.Server{Addr: h.bindAddr, Handler: mux}
```
masa213f, line 43: I don't use `cybozu-go/well` to keep the log format consistent with the reconciler.
For this usage, the default http server is enough, in my opinion.

zoetrope: I agree. 👍

cmd/moco-controller/cmd/run.go
```diff
@@ -136,6 +137,13 @@ func subMain(ns, addr string, port int) error {
 		return err
 	}
 
+	if config.pprofAddr != "" {
```
zoetrope, line 140: I think it would be better to disable pprof even when `0` is specified.
This is because it will be incompatible when using controller-runtime's pprof in the future.

masa213f: That's good. Fixed. 6cf0e1c

kmdkuk (approved):
LGTM

zoetrope (approved):
LGTM

PR #508, opened 2023-02-16
Update fluent-bit v2.0.9

masa213f:
Update the default fluent-bit version to v2.0.9 from v1.9.7.

The major version of fluent-bit has gone up, but there are no significant changes to the plugins which MOCO uses.
https://github.com/fluent/fluent-bit/releases/tag/v2.0.0

> **Note**
> MOCO uses the tail input plugin and the file output plugin.
> https://github.com/cybozu-go/moco/blob/v0.14.1/controllers/mysqlcluster_controller.go#L523-L536

### DIFF
--- a/version.go
+++ b/version.go
@@ -5,7 +5,7 @@ const (
 	Version = "0.14.1"
 
 	// FluentBitImage is the image for slow-log sidecar container.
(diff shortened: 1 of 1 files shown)


kmdkuk (approved):
LGTM

PR #510, opened 2023-02-22
Wait until all Pods are deleted in E2E

d-kuro:
The machine resources available to GitHub Actions runners are limited,
and it has been confirmed that instability due to issues such as OOM can occur when multiple MySQL Cluster pods created for multiple E2E scenarios exist at the same time.
In this PR, we will wait for the pod to disappear during E2E cleanup to avoid exceeding the expected resource usage of the runner.


masa213f (approved):
LGTM

PR #516, opened 2023-03-10
Disable the delay check if .spec.maxDelaySeconds == 0

masa213f:
This PR fixes the behavior when `.spec.maxDelaySeconds` is `0`.
We made it possible to disable the delay check when  `.spec.maxDelaySeconds` is `0` in https://github.com/cybozu-go/moco/pull/434.
However, that change seemed to be incorrect, so I fixed it.

In the current implementation, if we set `.spec.maxDelaySeconds: 0` in the MySQLCluster, the default value (`60s`) is used as a delay.
Because MOCO (or kube-apiserver) cannot distinguish whether the `0` is the zero value of integer type or intentionally set.
This PR changes the type of `maxDelaySeconds` to the pointer of int. This will allow the MOCO to determine if the 0 is set intentionally or not.


umezawatakeshi:
Do we have to test the case of maxDelaySeconds == 0 and nil?

umezawatakeshi (approved):

PR #526, opened 2023-05-16
Disable innodb_undo_log_truncate

masa213f:
refs https://github.com/cybozu-go/moco/issues/530

This PR disables `innodb_undo_log_truncate` by default.
Because the `undo log truncate` always fails on replica instances which are `set super_read_only = 1`.

https://dev.mysql.com/doc/refman/8.0/en/innodb-parameters.html#sysvar_innodb_undo_log_truncate
https://bugs.mysql.com/bug.php?id=104573


toshipp:
It would be better to have a comment on why the change is necessary. If it is related to a bug, add a link.

ymmt2005 (approved):
LGTM

PR #529, opened 2023-05-17
Build mysqld-exporter container image.

d-kuro:
refs: https://github.com/cybozu-go/moco/issues/535

Push mysqld-exporter container image to ghcr.io.

This PR only pushes the image.
The pushed image needs to be changed to be used by MOCO, but this will be done in another PR.


masa213f (approved):
LGTM

Issue #524, opened 2023-04-24
Why `MySQLClusterSpec.backupPolicyName` is required ?


pddg:
`backupPolicyName` is marked as `Required: true` in the document.
https://cybozu-go.github.io/moco/crd_mysqlcluster_v1beta1.html#mysqlclusterspec
https://cybozu-go.github.io/moco/crd_mysqlcluster_v1beta2.html#mysqlclusterspec

It seems to be optional since the type of `backupPolicyName` is `*string`.
There were no problems when I actually used the `MySQLCluster` without `backupPolicyName`.

umezawatakeshi:
As you pointed out, it is a bug of documentation. We will fix it.

PR #532, opened 2023-05-30
MySQLClusterSpec.BackupPolicyName should be annotated omitempty

umezawatakeshi:
fix: https://github.com/cybozu-go/moco/issues/524


masa213f (approved):
LGTM

Issue #537, opened 2023-07-04
Support reduce volume size


d-kuro:
## What

Support for reducing volume size.

Currently, MOCO supports volume expansion, but not volume reduction.

However, there may be situations where users may want to reduce the volume size depending on their use cases.

## How

Proposals will be prepared.

## Checklist

- [x] Finish implementation of the issue
- [x] Test all functions
- [x] Have enough logs to trace activities
- [x] Notify developers of necessary actions

PR #538, opened 2023-07-04
Add proposal to reduce volume size

d-kuro:
refs: https://github.com/cybozu-go/moco/issues/537

### DIFF
--- a/docs/designdoc/support_reduce_volume_size.md
+++ b/docs/designdoc/support_reduce_volume_size.md
@@ -1,0 +1,133 @@
+# Support for Reducing Volume Size
+
+## Context
+
+Support for reducing volume size.
+
+Currently, MOCO supports [volume expansion](./support_apply_pvc_template_changes.md), but not volume reduction.
+The reduction of the size of a Persistent Volume (PV) or Persistent Volume Claim (PVC) is not endorsed as a fundamental principle of Kubernetes.
+
+However, there may be situations where users may want to reduce the volume size depending on their use cases.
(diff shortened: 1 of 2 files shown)


docs/designdoc/support_reduce_volume_size.md
```diff
@@ -0,0 +1,153 @@
...
+
+```text
+moco_cluster_statefulset_recreate_total{name="mycluster", namespace="mynamesapce"} 3
+moco_cluster_statefulset_recreate_errors_total{name="mycluster", namespace="mynamesapce"} 1
+```
+
+If a StatefulSet fails to recreate, the metrics in `moco_cluster_statefulset_recreate_errors_total` is incremented after each reconcile,
+so users can notice anomalies by monitoring this metrics.
+
+### Conditions
```
masa213f, line 135: As commented verbally, this status is not necessary.

In the volume reduction steps, the scope of moco's responsibility is only step 2 (update StatefulSet).
Other steps (deleting pods, etc.) are the user's responsibility.
So there is no need to manage it as status.

masa213f (changes requested):

docs/designdoc/support_reduce_volume_size.md
d-kuro: Thanks for the feedback, I'll try to get it back to you.
I believe the description of the condition is gone now, so please review it again 🙏

masa213f (approved):
LGTM

Issue #535, opened 2023-06-07
Migrate image registry to qury.io to ghcr.io 


d-kuro:
## What

Migrate image registry to qury.io to ghcr.io.

Change the container images used by MOCO that are currently managed by neco-containers to be managed in this repository.
Use ghcr.io for container registry.
https://github.com/cybozu/neco-containers

## Checklist

- [x] Finish implementation of the issue
- [x] Test all functions
- [x] Have enough logs to trace activities
- [x] Notify developers of necessary actions

PR #542, opened 2023-07-22
Use image hosted on ghcr.io

d-kuro:
* https://github.com/cybozu-go/moco/issues/535

Use image hosted on `ghcr.io`.

We also use a local mysql image to run E2E tests after building the MySQL image.
This allows more detailed testing of the built mysql image.

### DIFF
--- a/.github/workflows/build-mysql-container.yaml
+++ b/.github/workflows/build-mysql-container.yaml
@@ -36,12 +36,7 @@ jobs:
 
           done
 
-          if [ "${#array[@]}" -eq 0 ]; then
-            json_output="[]"
-          else
-            json_output=$(printf '%s\n' "${array[@]}" | jq -R . | jq -s . | jq -c .)
-          fi
-
+          json_output=$(echo "${array[@]}" | jq -Rc 'split(" ")')
           echo "GITHUB_OUTPUT: mysql-versions=$json_output"
           echo "mysql-versions=$json_output" >> "$GITHUB_OUTPUT"
 
--- a/.github/workflows/build-mysql-container.yaml
+++ b/.github/workflows/build-mysql-container.yaml
@@ -15,15 +15,50 @@ on:
       - "!**.md"
 
 jobs:
+  filter:
+    runs-on: ubuntu-20.04
+    outputs:
+      mysql-versions: ${{ steps.filter.outputs.mysql-versions }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: filter
+        id: filter
+        working-directory: containers
+        run: |
+          array=()
+
+          for dir in $(find ./mysql -mindepth 1 -type d -printf '%f\n'); do
+            result=$(./tag_exists moco/mysql "mysql/$dir")
+
+            if [ "$result" = 'ng' ]; then
+              array+=( "$dir" )
+            fi
+
+          done
+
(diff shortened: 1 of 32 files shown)


.github/workflows/build-mysql-container.yaml
```diff
@@ -15,15 +15,50 @@ on:
...
+          done
+
+          if [ "${#array[@]}" -eq 0 ]; then
+            json_output="[]"
+          else
+            json_output=$(printf '%s\n' "${array[@]}" | jq -R . | jq -s . | jq -c .)
+          fi
+
+          echo "GITHUB_OUTPUT: mysql-versions=$json_output"
+          echo "mysql-versions=$json_output" >> "$GITHUB_OUTPUT"
```
d-kuro, line 46: Previously, Step "Check TAG file" would fail the job if the image already existed.
This caused a bug that caused GitHub Actions jobs to fail if the tag file contained some unchanged mysql versions.

I added filter jobs to build only mysql images that have changed

d-kuro: I tested jobs:

refs: https://github.com/cybozu-go/moco/actions/runs/5664698208?pr=542

masa213f: Good changes!

So don't we need to run `tag_exists` in the `Check TAG file` step? Or it is better to leave it?

masa213f (changes requested):

.github/workflows/build-mysql-container.yaml
d-kuro: Thank you very much!You're right, the double tag file check was unnecessary, so I removed the duplicate process.
https://github.com/cybozu-go/moco/pull/542/commits/70415756959286fc0de1c6825add6f6a66074ca4

.github/workflows/build-mysql-container.yaml
```diff
@@ -56,54 +91,34 @@ jobs:
         run: |
           container-structure-test test --image ghcr.io/cybozu-go/moco/mysql:${{ matrix.mysql-version }} --config ./containers/mysql/${{ matrix.mysql-version }}/container-structure-test.yaml
 
-      - name: Create kind cluster
-        run: kind create cluster --name=moco --config=./e2e/kind-config_actions.yaml --image=kindest/node:v${{ matrix.k8s-version }} --wait 1m
+      - run: |
+          swapon > swapon.txt
+          sudo swapoff -a
+          cat swapon.txt | tail -n+2 | awk '$2=="file" {print $1}' | sudo xargs --no-run-if-empty rm
+      - run: sudo mkdir /mnt/local-path-provisioner0 /mnt/local-path-provisioner1 /mnt/local-path-provisioner2
```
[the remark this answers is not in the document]
d-kuro: 📝 I'm transcribing my notes from a verbal conversation here.

The modifications extend beyond the scope of this Pull Reqeust, so I'd like to split it up into a separate Pull Reqeust.
Fix swap and mount settings to be switched at make file runtime using flags, etc.

masa213f (dismissed):
LGTM

masa213f:
I rebased this PR because it conflicted with https://github.com/cybozu-go/moco/pull/548.

masa213f (approved):

PR #552, opened 2023-08-09
Support redunce volume size

d-kuro:
https://github.com/cybozu-go/moco/issues/537

Support redunce volume size in MOCO.
It is implemented based on the proposals added below.

* https://github.com/cybozu-go/moco/pull/538

### DIFF
--- a/controllers/pvc.go
+++ b/controllers/pvc.go
@@ -131,6 +131,9 @@ func (r *MySQLClusterReconciler) resizePVCs(ctx context.Context, cluster *mocov1
 		case i == 0: // volume size is equal
 			continue
 		case i == 1: // current volume size is greater than new size
+			// The size of the Persistent Volume Claims (PVC) cannot be reduced.
+			// Although MOCO permits the reduction of PVC, an automatic resize is not performed.
+			// An error arises if a PVC involving reduction is passed, as it's unexpected.
 			return resizedPVC, fmt.Errorf("failed to resize pvc %q, want size: %s, deployed size: %s: %w", pvc.Name, newSize.String(), pvc.Spec.Resources.Requests.Storage().String(), ErrReduceVolumeSize)
 		case i == -1: // current volume size is smaller than new size
 			pvc.Spec.Resources.Requests[corev1.ResourceStorage] = *newSize
--- a/controllers/pvc.go
+++ b/controllers/pvc.go
@@ -174,7 +174,12 @@ func (*MySQLClusterReconciler) needResizePVC(cluster *mocov1beta2.MySQLCluster, 
 		case i == 0: // volume size is equal
 			continue
 		case i == 1: // volume size is greater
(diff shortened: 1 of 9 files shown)


e2e/pvc_test.go
```diff
@@ -24,6 +21,9 @@ var pvcTestYAML string
 var pvcApplyYAML string
 
 var _ = Context("pvc_test", func() {
+	if doUpgrade {
+		return
+	}
```
d-kuro, line 26: This scenario was not originally required to be run in the upgrade test, so we have changed it to skip the entire test.

As in the following test:
https://github.com/cybozu-go/moco/blob/b63186fc492f206813e1684abaecbbc0b4ff8c1e/e2e/lifecycle_test.go#L24

controllers/pvc.go
```diff
@@ -174,7 +174,12 @@ func (*MySQLClusterReconciler) needResizePVC(cluster *mocov1beta2.MySQLCluster,
 		case i == 0: // volume size is equal
 			continue
 		case i == 1: // volume size is greater
-			return nil, false, fmt.Errorf("failed to resize pvc %q, want size: %s, deployed size: %s: %w", pvc.Name, wantSize, deployedSize, ErrReduceVolumeSize)
```
masa213f, line 177: Please update the following test in `controllers/pvc_test.go`.
https://github.com/cybozu-go/moco/blob/v0.16.1/controllers/pvc_test.go#L157-L163

Also, add a new test case. 
A MySQLCluster has 2 volume claim templates. One expands the volume size, and the other reduces it.

e2e/pvc_test.go
```diff
@@ -66,107 +67,71 @@ var _ = Context("pvc_test", func() {
...
+	})
 
-			var pvcList corev1.PersistentVolumeClaimList
-			if err := json.Unmarshal(out, &pvcList); err != nil {
-				return err
-			}
-			if len(pvcList.Items) < 1 {
-				return errors.New("not found pvcs")
-			}
+	It("should volume size reduce succeed", func() {
```
masa213f, line 102: Can you confirm that the data is not lost after this test?

masa213f (changes requested):

e2e/pvc_test.go
d-kuro: Thanks, I added test to check for data loss.
https://github.com/cybozu-go/moco/pull/552/commits/57652b63c8237fa2a2fe4eaf2a8086a73507d0b1

controllers/pvc.go
d-kuro: I added tests.
Also, there was a problem with the test code, which has been corrected.
https://github.com/cybozu-go/moco/pull/552/commits/32b8fbbd9cc1b4ea2b024e04f6f1a82a57282d2f

masa213f:
LGTM. Thank you!

masa213f (approved):
LGTM. Thank you!

PR #561, opened 2023-08-29
Kill old connections when demoting primary

masa213f:
This PR fixes failover failure.

When the network of a node that the Primary MySQL Pod runs on is unavailable, the MOCO executes a failover.
However, if the network of the node is restored during the failover, 
failover gets stuck, and MySQLCluster will be unavailable for a long time.

### DIFF
--- a/clustering/manager_test.go
+++ b/clustering/manager_test.go
@@ -583,7 +583,147 @@ var _ = Describe("manager", func() {
 		}
 	})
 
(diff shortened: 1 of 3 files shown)


ymmt2005 (commented):
Can you add a test?

ymmt2005 (approved):
LGTM

zoetrope (approved):
LGTM

PR #565, opened 2023-09-06
backup: PITR did not work sometimes

### LINKED ISSUE #563: Point-in-Time Recovery feature does not apply binlog
**Describe the bug**

We are developing a "backup verifier" to verify the integrity of backups taken by MOCO (required by auditors :)). During the development of the backup verifier, we suspected that MOCO's point-in-time recovery feature was not applying the binlog that should be applied. The scenario we ran was as follows:

- Set up a CronJob that reports gtid_executed on MySQL every hour at 59 minutes
- Ran a backup job at 2023-09-02T03:03:00Z
- Ran another backup job at 2023-09-03T03:03:00Z
- Restored MySQL from backup with restorePoint: 2023-09-03T00:00:00Z

Since the reporting gtid_executed is executed every hour around 59 minutes, the restored MySQL should contain the GTID reported at 2023-09-02T23:59:00Z or 2023-09-02T22:59:00Z. However the restored MySQL containes GTIDs as of 2023-09-02T02:59:00Z.

This suggests that MOCO's restore process successfully restored from the dump.tar at 2023-09-02T03:03:00Z, but skipped applying the binlog from 2023-09-02T03:03:00Z to 2023-09-03T00:00:00Z.

In addition, when we checked the log of the restore job, we found the following messages:

```text
{"level":"info","ts":"2023-09-04T10:12:34Z","msg":"loaded dump successfully"}
{"level":"info","ts":"2023-09-04T10:12:34Z","msg":"applied binlog successfully"}
```

According to the source code, the following process (applyBinlog) should be executed between "loaded dump successfully" and "applied binlog successfully":

- Download binlog.tar.zst from object storage
- Extract binlog with zstd and tar
- `mysqlbinlog | mysql`

We manyally downloaded the binlog.tar.zst that should have been applied from object storage and decoded the binlog with mysqlbinlog. Decoding the binlog took more than 3 seconds. Considering that downloading from object storage and applying transactions to MySQL is required, applyBinlog should take at least a few seconds.
However, MOCO's restore job logs says that this process took less than one second. This suggests that the applyBinlog process was skipped at some point.

**Environments**
- Version: MOCO v0.16.1
- OS: Flatcar Container Linux (stable)

**To Reproduce**

1. Create MySQLCluster
2. Let `G1` be `SELECT @@gtid_executed`
3. Take a backup
4. Execute several transactions
5. Let `G2` be `SELECT @@gtid_executed`
6. Let `T` be `date --iso-8601=seconds`
7. Take a backup again
8. Restore MySQL with restore_point: `T`

**Expected behavior**

The restored MySQL contains `G2`.

ymmt2005:
A bug in RestoreManager.FindNearestDump caused the problem.
Fixes #563.

### DIFF
--- a/backup/restore.go
+++ b/backup/restore.go
@@ -138,7 +138,6 @@ func (rm *RestoreManager) Restore(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("failed to list object keys: %w", err)
 	}
-	sort.Strings(keys)
 
 	dumpKey, binlogKey, backupTime := rm.FindNearestDump(keys)
 	if dumpKey == "" {
@@ -195,16 +194,19 @@ func (rm *RestoreManager) Restore(ctx context.Context) error {
 	return nil
 }
 
+// FindNearestDump finds the nearest dump file and binlog file to the restore point.
+// `keys` are object keys for the restoring instance. They need not be sorted.
 func (rm *RestoreManager) FindNearestDump(keys []string) (string, string, time.Time) {
+	sort.Strings(keys)
+
 	var nearest time.Time
 	var nearestDump, nearestBinlog string
 
 	for _, key := range keys {
-		if strings.HasSuffix(key, constants.BinlogFilename) {
-			nearestBinlog = key
-			continue
-		}
-		if !strings.HasSuffix(key, constants.DumpFilename) {
+		isBinlog := strings.HasSuffix(key, constants.BinlogFilename)
+		isDump := strings.HasSuffix(key, constants.DumpFilename)
+		if !isBinlog && !isDump {
+			rm.log.Info("skipping garbage", "key", key)
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


arosh (approved):
LGTM

masa213f (approved):
LGTM

masa213f:
If there were some binlog files in the bucket, the MOCO always used to select the most recent binlog file, regardless of the time specified.
Therefore, MOCO was in danger of restoring incorrect data.

ymmt2005:
> If there were some binlog files in the bucket, the MOCO always used to select the most recent binlog file, regardless of the time specified.

If there's a newer binlog file than the one we should use, MOCO wrongly picked the newer one.

> Therefore, MOCO was in danger of restoring incorrect data.

Not exactly. MOCO didn't apply any transaction in the bad case,
so it only restores the full dump data without applying binlogs.

masa213f:
@ymmt2005 Thank you. I understood.

PR #568, opened 2023-09-11
Fix image name and tag

masa213f:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,7 @@ USER 10000:10000
 ENTRYPOINT ["/moco-controller"]
 
 # For MySQL binaries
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -20,7 +20,7 @@ USER 10000:10000
 ENTRYPOINT ["/moco-controller"]
 
 # For MySQL binaries
-FROM --platform=$TARGETPLATFORM ghcr.io/cybozu-go/moco/mysql:8.0.32.1 as mysql
```
masa213f, line 23: We have deleted images that had the incorrect tag.
So, the `ghcr.io/cybozu-go/moco/mysql:8.0.32.1` does not exist.

umezawatakeshi (approved):

PR #571, opened 2023-09-12
Build moco-controller with go 1.21

yamatcha:
Upgrade go version to 1.21

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build the moco-controller binary
-FROM --platform=$BUILDPLATFORM quay.io/cybozu/golang:1.19-focal as builder
+FROM --platform=$BUILDPLATFORM ghcr.io/cybozu/golang:1.21-jammy as builder
 
 ARG TARGETARCH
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build the moco-controller binary
-FROM --platform=$BUILDPLATFORM quay.io/cybozu/golang:1.19-focal as builder
+FROM --platform=$BUILDPLATFORM ghcr.io/cybozu/golang:1.21-jammy as builder
```
masa213f, line 2: Please use the same varion of Ubuntu for the build stage, mysql:8.0.32.2 and the backup image.

```
# For MySQL binaries
FROM --platform=$TARGETPLATFORM ghcr.io/cybozu-go/moco/mysql:8.0.32.2 as mysql

# the backup image
FROM --platform=$TARGETPLATFORM quay.io/cybozu/ubuntu:20.04
```

masa213f (changes requested):

masa213f (approved):

PR #584, opened 2023-10-18
Use Gomega matcher in Eventually and check pods' labels

masa213f:
This PR makes the following changes to the envtest.

1. Use Gomega matcher in `Eventually` block.
    It's just a refactoring.
2. Check the pods' metadata in `Eventually` or `Consistently` blocks.
    The pods' metadata are updated asynchronously with MySQLCluster's `.status`.
    With the current implementation, pods's metadata are checked just once using `Expect`.
    Therefore, if there is a delay in updating the pods' metadata, the test may fail. 
3. Add more checks for pods's labels.
    Preparation for https://github.com/cybozu-go/moco/pull/561

### DIFF
--- a/clustering/manager_test.go
+++ b/clustering/manager_test.go
@@ -3,7 +3,6 @@ package clustering
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"os"
 	"time"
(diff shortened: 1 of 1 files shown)


zoetrope (approved):
LGTM

PR #587, opened 2023-10-20
Kill existing connections when changing roles

masa213f:
This PR resolve the issue https://github.com/cybozu-go/moco/issues/597
I tried to implement e2e but could not.

### DIFF
--- a/clustering/status.go
+++ b/clustering/status.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
@@ -14,7 +15,6 @@ import (
 	"github.com/cybozu-go/moco/pkg/constants"
 	"github.com/cybozu-go/moco/pkg/dbop"
 	"github.com/cybozu-go/moco/pkg/password"
-	"golang.org/x/exp/slices"
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
--- a/clustering/status.go
+++ b/clustering/status.go
@@ -14,6 +14,7 @@ import (
 	"github.com/cybozu-go/moco/pkg/constants"
 	"github.com/cybozu-go/moco/pkg/dbop"
 	"github.com/cybozu-go/moco/pkg/password"
+	"golang.org/x/exp/slices"
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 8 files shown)


clustering/status.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/cybozu-go/moco/pkg/constants"
 	"github.com/cybozu-go/moco/pkg/dbop"
 	"github.com/cybozu-go/moco/pkg/password"
+	"golang.org/x/exp/slices"
```
ymmt2005, line 17: Can't we use the standard package?
https://pkg.go.dev/slices

docs/clustering.md
```diff
@@ -249,7 +249,6 @@ There is nothing can be done.
 - Stop replication of errant replicas.
 - Set `super_read_only=1` for replica instances that are writable.
 - Adjust `moco.cybozu.com/role` label to Pods according to their roles.
-    - For errant replicas, the label is removed to prevent users from reading inconsistent data.
```
ymmt2005, line 252: Why do you remove this specification?

clustering/status.go
masa213f: Thank you! I don't know the standard package. I'll use it.

docs/clustering.md
masa213f: I thought it would be easier to determine which instance is in progress changing roles.
Should I keep this specification?

masa213f: reverted 190f834fa99421daece8c99ac110099d496eb666

clustering/status.go
masa213f: fixed 3b9fa2c940287bf1396bce195f541ebce2ad2b48

docs/clustering.md
ymmt2005: Actually, preventing users from accessing errant replicas is more important IMO.

ymmt2005: ah, I understood your intention.
You wanted to add `errant-replica` as a role to indicate that the instance is neither primary nor a replica.

ymmt2005: > For errant replicas, the label is removed to prevent users from reading inconsistent data.

How about adding this line instead?

```
- For errant replicas, the label is set to `errant-replica`, which effectively prevents clients from connecting them via a Service.
```

masa213f: @ymmt2005 
Thanks for the advice.
Yes. I was trying to add `moco.cybozu.com/role: errant-replica` label to errant replicas and to kill the connections on the errants just in case.

However, I think killing the errant replica's connections is almost ineffective.
And I could fix the issue https://github.com/cybozu-go/moco/issues/597 without the breaking change.
So, I will not change the specification of the role label.

ymmt2005: OK.

ymmt2005 (approved):
LGTM

Issue #545, opened 2023-07-25
Remove v1beta1 API


masa213f:
## What

Time has passed since the release of `v1beta2` API on [MOCO v0.11.0](https://github.com/cybozu-go/moco/releases/tag/v0.11.0)
It should be time to delete `v1beta1` API.

## How

We need at least the following three releases to delete old API safety.
ref: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#overview

1. Change the storage version.
    - Set `storage: true` to `v1beta2` API.
    - Set `storage: false` and `deprecated: true` to `v1beta1` API. 
    - Update API version in ownerReferences in child resources created by `v1beta1` API.
    - Announce to upgrade existing objects to a new stored version.
        - https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#upgrade-existing-objects-to-a-new-stored-version
2. Disable the `v1beta1` API.
    - Set `served: false` to `v1beta1` API. 
3. Delete the `v1beta1` API.
    - Delete the `v1beta1` API and conversion webhook.

## Checklist

- [ ] Finish implementation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

PR #592, opened 2023-10-25
api: add deprecatedversion marker

shunki-fujita:
add deprecatedversion marker.
Part of https://github.com/cybozu-go/moco/issues/545

### DIFF
--- a/api/v1beta1/backuppolicy_types.go
+++ b/api/v1beta1/backuppolicy_types.go
@@ -83,6 +83,7 @@ type BackupPolicySpec struct {
 }
 
 //+kubebuilder:object:root=true
+//+kubebuilder:deprecatedversion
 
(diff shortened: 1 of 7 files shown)


ymmt2005 (commented):
Question.
I know the `warning` is optional, but what happens if we don't give it?

shunki-fujita:
@ymmt2005 
> Question. I know the `warning` is optional, but what happens if we don't give it?

Probably use the following format:
https://github.com/kubernetes/kubernetes/blob/558a6797a2962dd77c38345549d2a2452bbc3cdc/staging/src/k8s.io/apiextensions-apiserver/pkg/apiserver/customresource_handler.go#L1037

ymmt2005:
Thanks.

ymmt2005 (approved):
LGTM

PR #605, opened 2023-11-10
pass password via command arguments on backup

umezawatakeshi:
If `--passwords-from-stdin` is used, password prompt will be output and it may be handled as a log line without LF. The behavior is problematic because the log line occupies the partial line buffer in logging agent like promtail and will never be flushed naturally.

The command arguments are not visible from outside of the Pod. So this modification is not insecure.

### DIFF
--- a/pkg/bkop/backup.go
+++ b/pkg/bkop/backup.go
@@ -13,7 +13,7 @@ import (
 func (o operator) DumpFull(ctx context.Context, dir string) error {
 	args := []string{
 		fmt.Sprintf("mysql://%s@%s:%d", o.user, o.host, o.port),
-		"--passwords-from-stdin",
+		"-p" + o.password,
 		"--save-passwords=never",
 		"-C", "False",
 		"--",
@@ -25,7 +25,6 @@ func (o operator) DumpFull(ctx context.Context, dir string) error {
 	}
 
 	cmd := exec.CommandContext(ctx, "mysqlsh", args...)
-	cmd.Stdin = strings.NewReader(o.password)
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 	return cmd.Run()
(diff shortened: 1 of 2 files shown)


kmdkuk (approved):

pkg/bkop/backup.go
```diff
@@ -13,7 +13,7 @@ import (
 func (o operator) DumpFull(ctx context.Context, dir string) error {
 	args := []string{
 		fmt.Sprintf("mysql://%s@%s:%d", o.user, o.host, o.port),
-		"--passwords-from-stdin",
+		"-p" + o.password,
```
ymmt2005, line 16: In this case, I'd suggest embedding the password in the URI.
```go
		fmt.Sprintf("mysql://%s:%s@%s:%d", o.user, o.password, o.host, o.port),
```
https://dev.mysql.com/doc/refman/8.0/en/connecting-using-uri-or-key-value-pairs.html

pkg/bkop/restore.go
```diff
@@ -23,7 +23,7 @@ func (o operator) PrepareRestore(ctx context.Context) error {
 func (o operator) LoadDump(ctx context.Context, dir string) error {
 	args := []string{
 		fmt.Sprintf("mysql://%s@%s:%d", o.user, o.host, o.port),
```
ymmt2005, line 25: ditto

pkg/bkop/backup.go
umezawatakeshi: @ymmt2005 
I don't care either way but what's the pros of it?
cons: password should be escaped like:
```
	u := url.URL{
		Scheme: "mysql",
		Host:   fmt.Sprintf("%s:%d", o.host, o.port),
		User:   url.UserPassword(o.user, o.password),
	}
	args := []string{
		u.String(),
                ...
```

ymmt2005: @umezawatakeshi Fair enough.
Then, I'd withdraw my suggestion.

ymmt2005: I'm afraid we should've escaped the user name in the URI.

umezawatakeshi: @ymmt2005 
It seems that [constants](https://github.com/cybozu-go/moco/blob/v0.18.1/pkg/constants/users.go#L3-L13) are used as usernames in moco-backup container and they cannot be passed from outside. So, it's not necessary to escape username itself.

Issue #558, opened 2023-08-23
Update of `annotations` field in volumeClaimTemplate is not reflected in `annotations` field in existing PersistentVolumeClaims


arosh:
**Describe the bug**
We use MOCO with [pvc-autoresizer](https://github.com/topolvm/pvc-autoresizer). This is because we want to automatically resize the volumes created by MOCO in response to lack of free space by setting the annotations defined by pvc-autoresizer in MySQLCluster's volumeClaimTemplate. These annotations allow us to configure the resizing behavior, such as thresholds. Our team has adopted a change in resizing policy, so we updated the annotations of MySQLCluster's volumeClaimTemplate to apply the new policy. We expected that the updated annotations of the volumeClaimTemplate to be reflected in the annotations of the already created PersistentVolumeClaims, but unexpectedly the annotations of PersistentVolumeClaim remained unchanged.

**Environments**
- Version: MOCO v0.16.1, [pvc-autoresizer](https://github.com/topolvm/pvc-autoresizer) v0.8.1
- OS: Flatcar Container Linux (stable)

**To Reproduce**
1. Configure annotations of volumeClaimTemplate as the follows and deploy MySQLCluster

```
resize.topolvm.io/increase: 5Gi
resize.topolvm.io/storage_limit: 100Gi
resize.topolvm.io/threshold: 5Gi
```

2. Update annotations of volumeClaimTemplate as the follows.

```
resize.topolvm.io/increase: 5Gi
resize.topolvm.io/storage_limit: 200Gi
resize.topolvm.io/threshold: 5Gi
```

3. Check `kubectl describe pvc mysql-data-moco-...`

**Expected behavior**
The updated annotations of the volumeClaimTemplate to be reflected in the annotations of the already created PersistentVolumeClaims.

masa213f:
It's by design. Please update the metadata by any operation.

The metadata of PVCs may be updated by someone other than moco, like admission webhooks.
Because of the possibility of malfunction, MOCO will not change labels and annotations of existing PVCs.

https://github.com/cybozu-go/moco/blob/v0.16.1/docs/change-pvc-template.md
> When re-creating a StatefulSet, moco-controller does not support any operation other than volume expansion as described below. It only re-creates the StatefulSet. Therefore, changes to the label and annotation assigned to the PVC must be done by the user. This is because there is a concern that if someone other than the moco-controller is editing the PVC's metadata, there may be side effects.

masa213f:
[...]

PR #613, opened 2023-11-29
Sync PVC labels and annotations

d-kuro:
refs: https://github.com/cybozu-go/moco/issues/558

https://github.com/cybozu-go/moco/issues/558#issuecomment-1690940455

Update the value if the label and annotation of the PVC is identical to the label and annotation key of the volumeClaimTemplate based on the above comment specification.

### DIFF
--- a/cmd/moco-controller/cmd/root.go
+++ b/cmd/moco-controller/cmd/root.go
@@ -34,6 +34,8 @@ var config struct {
 	backupImage             string
 	fluentBitImage          string
 	exporterImage           string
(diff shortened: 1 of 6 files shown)


d-kuro:
@kmdkuk
Thank you for the comment.
I fixed the implementation to sync annotations and labels to PVC based on the key set in the controller's flag.

kmdkuk (approved):
LGTM

zoetrope (approved):

PR #619, opened 2023-12-04
Enable cancel-in-progress in ci

yamatcha:


.github/workflows/ci.yaml
```diff
@@ -6,6 +6,10 @@ on:
     branches:
       - 'main'
 
+concurrency:
+  group: ${{ github.head_ref || github.run_id }}
```
ymmt2005, line 10: Just a suggestion
```suggestion
  group: ${{ github.head_ref || github.run_id }}-ci
```

ymmt2005: In the official reference, they actually use `${{ github.workflow }}` to specify the workflow.

yamatcha: Thank you. I've fixed as you suggested.

ymmt2005: I mean, if we use `${{ github.workflow }}`, we don't need to add `-e2e` or `-ci`.

ymmt2005 (approved):
LGTM

PR #637, opened 2024-01-22
Support backup and restore with ObjectBucketName set in environment variables

yamatcha:
Currently, there is no way to set the `ObjectBucketName` in ConfigMap or Secrets. This can be accomplished by using envFrom without adding new functionality.
So we have added docs and some test cases in this PR.

Signed-off-by: yamatcha <soju-yamashita@cybozu.co.jp>

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -457,6 +457,35 @@ spec:
 ...
 ```
 
+>**Note:** If you want to specify the ObjectBucket name in a ConfigMap or Secret, you can use `envFrom` and specify the environment variable name in `jobConfig.bucketConfig.bucketName` as follows.
+>This behavior is tested.
+
+```yaml
+apiVersion: moco.cybozu.com/v1beta2
+kind: BackupPolicy
+metadata:
+  namespace: backup
+  name: daily
+spec:
+  jobConfig:
+    bucketConfig:
+      bucketName: "$(BUCKET_NAME)"
+      endpointURL: http://minio.default.svc:9000
+      usePathStyle: true
+    envFrom:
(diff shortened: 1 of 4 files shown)


docs/usage.md
```diff
@@ -457,6 +457,27 @@ spec:
 ...
 ```
 
+>**Note:** If you want to specify the ObjectBucket name in configMap or secret, you can use envFrom and specify environments variable in `ObjectBucketName` like below.
```
ymmt2005, line 460: Do you mean the below?
```suggestion
>**Note:** If you want to specify the ObjectBucket name in a ConfigMap or Secret, you can use `envFrom` and specify the environment variable name in `jobConfig.bucketConfig.bucketName` as follows.
```

docs/usage.md
```diff
@@ -457,6 +457,27 @@ spec:
...
+spec:
+  jobConfig:
+    bucketConfig:
+      bucketName: "$(BUCKET_NAME)"
+      endpointURL: http://minio.default.svc:9000
+      usePathStyle: true
+    envFrom:
+    - configMapRef:
+        name: bucket-name
+...
```
ymmt2005, line 478: It'd be more helpful if we include an example ConfigMap.
```yaml
...
---
apiVersion: v1
kind: ConfigMap
data:
  BUCKET_NAME: foobar

docs/usage.md
yamatcha: @ymmt2005 
Yes, thank you for pointing that out.

kmdkuk (approved):

Issue #647, opened 2024-02-09
A backup of another sourcename with the same prefix is restored.


yoheinbb:
**Describe the bug**
If there are backups with sourcename xx-1 and xx-11, restoring by specifying xx-1 and after the **last date and time** of the xx-1 backup will restore xx-11.

**Expected behavior**
If sourcename is xx-1, only the backup of xx-1 is restored.

**Additional context**
Backup paths xx-1 and xx-11 are both listed because the object is retrieved by prefix.
https://github.com/cybozu-go/moco/blob/1e5089dea41fa1127210ed7cba70f6f6070a1d60/pkg/bucket/s3.go#L112-L120

morimoto-cybozu:
Resolved by https://github.com/cybozu-go/moco/pull/648.

PR #648, opened 2024-02-14
Add / delimiter to the end of path in calcPrefix

yoheinbb:
refs:
#647

### DIFF
--- a/pkg/bucket/interface.go
+++ b/pkg/bucket/interface.go
@@ -14,5 +14,7 @@ type Bucket interface {
 	Get(ctx context.Context, key string) (io.ReadCloser, error)
 
 	// List lists the matching object keys that have `prefix`.
+	// The prefix argument must end in /. (e.g. "foo/bar/").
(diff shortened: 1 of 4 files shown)


pkg/bucket/interface.go
```diff
@@ -14,5 +14,7 @@ type Bucket interface {
 	Get(ctx context.Context, key string) (io.ReadCloser, error)
 
 	// List lists the matching object keys that have `prefix`.
+	// The prefix argument must end in /. (e.g. "foo/bar/").
```
ymmt2005, line 17: nit.
```suggestion
	// The prefix argument should end with /. (e.g. "foo/bar/").
```

ymmt2005 (approved):
LGTM. Thanks.

pkg/bucket/interface.go
yoheinbb: Thanks! I fixed it.
6f6bf886409b0f067572f066cf2aa0beb40086ef

ymmt2005 (approved):

yamatcha (approved):

PR #652, opened 2024-03-11
Use changed-files action instead of paths-ignore

yamatcha:
Use changed-files action instead of `paths-ignore` to pass `finish-e2e`  job(required job) when only docs or container are changed.
And add composite actions of e2e and dbtest and use those action.

Signed-off-by: yamatcha <soju-yamashita@cybozu.co.jp>


umezawatakeshi:
What benefit by this?

.github/workflows/ci-e2e.yaml
```diff
@@ -67,20 +102,33 @@ jobs:
...
       - run: make logs
         working-directory: e2e
-        if: always()
+        if: ${{steps.changed-files.outputs.any_changed}} == 'true' && always()
       - uses: actions/upload-artifact@v3
-        if: always()
+        if: ${{steps.changed-files.outputs.any_changed}} == 'true' && always()
         with:
           name: logs-upgrade.tar.gz
           path: e2e/logs.tar.gz
```
umezawatakeshi, line 134: How about splitting this section to a separate action file and use it ont only from this workflow but also from weekly.yaml ?
It makes job definition simpler.

umezawatakeshi (changes requested):

umezawatakeshi (approved):

PR #656, opened 2024-03-14
Fix specifying paths of files_ignore in ci-e2e workflow

yamatcha:


.github/workflows/ci-e2e.yaml
```diff
@@ -33,7 +33,7 @@ jobs:
         id: changed-files
         with:
           files_ignore: |
-            **.md
+            **/**.md
```
ymmt2005, line 36: `**/*.md` should be good.

PR #661, opened 2024-04-03
chart: add value to set a PriorityClass

npdgm:
Hi,
This will allow setting a custom PriorityClass for the controller, so that it run with a higher than normal priority and better resist to outages and busy clusters.
Cheers


zoetrope (approved):
LGTM

PR #662, opened 2024-04-13
 Option to to use localhost instead of pod name

vsliouniaev:
When running with istio enabled the communication between the agent and mysqld fails. To overcome this the admin interface for mysqld should be bound to `localhost` and the agent should communicate on `localhost` as well, rather than using pod name.

[More context here](https://github.com/cybozu-go/moco/pull/658#issuecomment-2012142633)

Corresponding moco-agent PR here: https://github.com/cybozu-go/moco-agent/pull/92

I also noticed that `max-delay` and `log-rotation-schedule` are being split across multiple lines when passed to the agent sidecar, which seemes unintentional.

### DIFF
--- a/api/v1beta2/mysqlcluster_types.go
+++ b/api/v1beta2/mysqlcluster_types.go
@@ -117,6 +117,11 @@ type MySQLClusterSpec struct {
 	// If set to true, the sidecar container is not added. The default is false.
 	// +optional
 	DisableSlowQueryLogContainer bool `json:"disableSlowQueryLogContainer,omitempty"`
+
+	// MySQLDLocalHost configures mysqld interface to bind and be accessed over localhost instead of pod name.
+	// During container init moco-agent will set mysql admin interface is bound to localhost. The moco-agent will also
+	// communicate with mysqld over localhost when acting as a sidecar.
+	MySQLDLocalHost bool `json:"mysqldLocalHost,omitempty"`
 }
 
 func (s MySQLClusterSpec) validateCreate() (admission.Warnings, field.ErrorList) {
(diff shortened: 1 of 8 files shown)


api/v1beta2/mysqlcluster_types.go
```diff
@@ -117,6 +117,11 @@ type MySQLClusterSpec struct {
 	// If set to true, the sidecar container is not added. The default is false.
 	// +optional
 	DisableSlowQueryLogContainer bool `json:"disableSlowQueryLogContainer,omitempty"`
+
+	// MySQLDLocalHost configures mysqld interface to bind and be accessed over localhost instead of pod name.
+	// During container init moco-agent will set mysql admin interface is bound to localhost. The moco-agent will also
+	// communicate with mysqld over localhost when acting as a sidecar.
+	MySQLDLocalHost bool `json:"mysqldLocalHost,omitempty"`
```
zoetrope, line 124: I apologize for the delayed response.

I find the option name `mysqldlocalHost` a bit unclear.
Perhaps something like `useLocalhost` or `agentUseLocalhost` might be clearer?
Of course, if there is a better suggestion, I'm open to that as well.

Since changes to the CRD are not trivial, I want to proceed with caution.

zoetrope (changes requested):

api/v1beta2/mysqlcluster_types.go
vsliouniaev: No problem, will update it to `agentUseLocalhost`

zoetrope (dismissed):
LGTM

zoetrope (dismissed):

zoetrope:
Seems like the unit tests are failing somehow.

vsliouniaev:
Believe I found the issue

vsliouniaev:
@zoetrope do I need to do anything else to get this merged? 🙏

zoetrope (approved):

Issue #638, opened 2024-01-22
binlog can be disabled


shunki-fujita:
**Describe the bug**
There is code that tries to disable binlog.
https://github.com/cybozu-go/moco/blob/a7352fa275f057ac6b1249eeeb6749d64b6504e3/pkg/mycnf/generator.go#L168

However, this variable is the filename of binlog and does not disable binlog.
https://dev.mysql.com/doc/refman/8.0/en/replication-options-binary-log.html#option_mysqld_log-bin

To disable binlog, either skip-log-bin or disable-log-bin must be specified.
Thus, delete() here is not working as intended.

**Expected behavior**
Specifying skip-log-bin or disable-log-bin does not disable binlog

PR #678, opened 2024-05-19
Prevent disabling of binlog

d-kuro:
refs: https://github.com/cybozu-go/moco/issues/638

### DIFF
--- a/pkg/mycnf/generator.go
+++ b/pkg/mycnf/generator.go
@@ -167,6 +167,8 @@ func Generate(userConf map[string]string, memTotal int64) string {
 	delete(mysqldConf, opaqueKey)
 	delete(mysqldConf, "log_bin")
 	delete(mysqldConf, "log_error")
+	delete(mysqldConf, "skip_log_bin")
+	delete(mysqldConf, "disable_log_bin")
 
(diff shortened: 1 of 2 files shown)


ymmt2005 (approved):
LGTM. Thank you.

vsliouniaev:
I'm currently using the operator with single DB instances and explicitly disabling binary logging as I don't need replication but a very lightweight implementation.

Is it possible to maintain this behavior?

ymmt2005:
I'm afraid not.
Even a single cluster needs binlogs for MOCO because we allow users to add replicas later, which needs binlogs.

vsliouniaev:
What if there was a CRD level option that would allow for this, provided that the number of replicas is 1? That would be easier to validate.

ymmt2005:
@masa213f @d-kuro thoughts?

d-kuro:
@vsliouniaev @ymmt2005 
Can `_includ` be used?
https://cybozu-go.github.io/moco/usage.html?highlight=_inc#opaque-configuration

ymmt2005:
yup, that should work.

vsliouniaev:
Thanks that's very helpful!

Issue #659, opened 2024-03-21
Scaling cluster down to 0


vsliouniaev:
## What

I'm hoping to use the operator to provide test clusters for developers. Since the development cycle is not 24 hours it would be more cost-effective to shut down the instances when not in use to free up resources.

I can see that deleting the entire cluster wipes the root access credentials as well as all PVCs, which is not ideal for this setup.

## How

It would be convenient if the option to scale to 0 replicas was allowed by the operator. 

Alternatively a different option to suspend the cluster that's not tied to the replica count may be preferable if some other logic is tied to increasing the number of replicas.

## Checklist

- [ ] Finish implementation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

ymmt2005:
My gut feeling is like the following.

Add a mode of operation to the moco-controller so that it removes StatefulSet but keeps the PVCs and all other resources. When the user wants to restore the cluster, the moco-controller will just create the StatefulSet, which will reuse the PVCs.

Does it sound good?

vsliouniaev:
Sounds like that would work. Scaling the STS down to 0 would probably be equivalent.

vsliouniaev:
I've done a little local experimenting with and have it working reasonably well, but the controllers are throwing a lot of benign errors as a result. If this approach proves workable in my setup I'll clean up the implementation. My approach is to add another property to the spec, rather than setting the spec.replicas=0, since the replica count in the CRD has implications for clustering.

PR #683, opened 2024-05-29
Allow setting a cluster offline

vsliouniaev:
Fixes: https://github.com/cybozu-go/moco/issues/659

Hi! I'm looking for guidance on the general approach here.

Overall this functions fine in my in-cluster testing but I'm only using a small sub-set of functionality.

### DIFF
--- a/api/v1beta2/mysqlcluster_types.go
+++ b/api/v1beta2/mysqlcluster_types.go
@@ -122,6 +122,10 @@ type MySQLClusterSpec struct {
 	// During container init moco-agent will set mysql admin interface is bound to localhost. The moco-agent will also
 	// communicate with mysqld over localhost when acting as a sidecar.
 	AgentUseLocalhost bool `json:"agentUseLocalhost,omitempty"`
+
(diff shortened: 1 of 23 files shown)


morimoto-cybozu:
@vsliouniaev 
Hi! Thank you for committing!
Are you wondering if you can move this pull request forward for review?
Your commit seems very reasonable and proper to me. But it seems to lack test code for the new property. How about adding a new `It()` case to `controllers/mysqlcluster_controller_test.go`? The case of `It("should delete all related resources")` would be a useful reference.
Please let us know if you feel you need another type of help.

vsliouniaev:
Certainly, was looking for some initial feedback on the overall approach

morimoto-cybozu:
@vsliouniaev 
In terms of code modification approach, it seems good to add a new property for an offline cluster. And the motivation for configuring a cluster as "offline" seems reasonable to me.
In terms of review process approach, making a pull request is enough.

shunki-fujita (approved):

morimoto-cybozu (approved):

Issue #682, opened 2024-05-29
CRDs in Helm chart should not be part of templates


vholer:
## What

Currently, the CRDs for Moco are part of the templates (`charts/moco/**templates**/generated/crds/`) so they are always created on Helm chart install and (nearly) always deleted on Helm chart uninstall. If anybody uninstalls Helm chart by accident while there are some MySQLClusters, this might lead to the data loss - not immediately on uninstall due to finalizers not being able to proceed with stopped controller, but after Helm chart with controller is installed back again!!

Since CRDs are part of templates, Helm doesn't distinguish them from other objects. Because of that, one can't use Helm's option to skip CRDs installation (`helm install --skip-crds`) if handled different way or even show them. E.g.,

```
$ helm show crds moco/moco
$ 
```

## How

Support for CRDs in Helm is not great, but the common practice (as described in [documentation](https://helm.sh/docs/topics/charts/#custom-resource-definitions-crds)) is to place them into top-level directory `crds/` within a chart, not into `templates/...`. This way they would be treated as CRDs and they could be skipped on installation (`--skip-crds`) or displayed on terminal.

I.e., move from `charts/moco/templates/generated/crds/` into `charts/moco/crds/`.

## Checklist

- [ ] Finish implementation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

masa213f:
@vholer 
Thank you for the report. Sounds good.
Would you like to fix it on your end?

vholer:
I was trying to do the change, but found a templated parts in CRDs, which make the move to `crds/` impossible (CRDs in Helm are not templated, https://helm.sh/docs/topics/charts/#custom-resource-definitions-crds). So I came with a different approach in #684 , which at least improves current situation - installs the CRDs conditionally (so user can disable them in Helm chart and handle installation on their own if they want), but mainly never deletes them (to avoid data loss).

If the PR is merged, I think this issue can be closed (although not fixed the best way).

masa213f:
@vholer 
Thank you for adding the excellent feature!
I added caution about updating the chart. Please read on :)
https://github.com/cybozu-go/moco/pull/685

If there are no other concerns, I would close this issue. Is it ok?

vholer:
@masa213f LGTM

It's true that the case when upgrading from older version while having CRDs disabled would delete them (as part of cleanup of previous Helm chart version). Thanks for the catch, we can close this.

PR #684, opened 2024-05-30
Helm chart CRDs optional and not deleted on uninstall

vholer:
I was trying to deal with #682 to move CRDs from `templates/generated/crds/` to `crds/`, but I found out that one CRD has configured conversion webhook and certificate, which are deployment/environment specific:
https://github.com/cybozu-go/moco/blob/a683b05a3cc2692114743f4c3f71ab660d22109c/charts/moco/templates/generated/crds/moco_crds.yaml#L2073-L2074 Hmm... So I came with a different approach, which is used also in some other projects: leave it part of templates, but make it conditional and prevent Helm to remove it.

Similar approach can be seen in cert-manager or Rook, e.g.:
- https://github.com/rook/rook/blob/master/deploy/charts/rook-ceph/templates/resources.yaml#L1
- https://github.com/rook/rook/blob/master/deploy/charts/rook-ceph/templates/resources.yaml#L8

Changes:
- new crds.enabled Helm variable to create CRDs conditionally
- label CRDs with `helm.sh/resource-policy: keep` to avoid their delete on uninstall

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,8 +63,10 @@ manifests: controller-gen kustomize yq ## Generate WebhookConfiguration, Cluster
 	mkdir -p charts/moco/templates/generated/crds/
 	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
 	$(KUSTOMIZE) build config/crd -o config/crd/tests # Outputs static CRDs for use with Envtest.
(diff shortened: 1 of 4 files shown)


masa213f:
@vholer 
Thanks for the change. It looks good to me at first glance.

However, I have one point of concern about the `crds.enabled`.
There was a little worrisome warning in the rook's chart description.

https://github.com/rook/rook/blob/v1.14.5/deploy/charts/rook-ceph/values.yaml#L17
> **WARNING** Only set during first deployment. If later disabled the cluster may be DESTROYED.

Do you know the behavior if the `crds.enabled` is disabled later?

masa213f:
@vholer 
I see. Thanks for the confirmation! It looks fine.

masa213f (approved):
LGTM

PR #687, opened 2024-06-06
Add basic Helm chart values

vholer:
Covers configurable
- replica count
- image pull policy
- reference to secrets for image pulling from authenticated container registry

Nothing really super important, but it might come handy to have these options easily configurable.


masa213f (approved):
LGTM. Thank you!

masa213f (approved):

PR #714, opened 2024-07-02
Remove weekly workflow

yamatcha:
Now, the test target of the [weekly workflow](https://github.com/cybozu-go/moco/blob/main/.github/workflows/weekly.yaml) is the same as the test target of the [e2e-ci workflow](https://github.com/cybozu-go/moco/blob/main/.github/workflows/ci-e2e.yaml).


ymmt2005 (approved):
LGTM

Issue #708, opened 2024-06-27
Long wait for SetReadOnly to complete


shunki-fujita:
**Describe the bug**
SetReadOnly waits for a running DML.
Therefore, if it waits for a long time, deleteGracePeriodSeconds may be reached.
https://github.com/cybozu-go/moco/blob/c545ebbbab9e92aaf9b07c42263464d9faec50f9/clustering/operations.go#L154-L156

**Expected behavior**
If SetReadOnly does not complete in a short time, give up and run KillConnections.

shunki-fujita:
The problem is that the switching is not completed within the PreStopSeconds time.
Solution will be considered separately.

PR #717, opened 2024-07-03
issue-708: Long wait for SetReadOnly to complete

shunki-fujita:
https://github.com/cybozu-go/moco/issues/708

### DIFF
--- a/clustering/operations.go
+++ b/clustering/operations.go
@@ -151,12 +151,36 @@ func (p *managerProcess) switchover(ctx context.Context, ss *StatusSet) error {
 	log.Info("begin switchover the primary", "current", ss.Primary, "next", ss.Candidate)
 
 	pdb := ss.DBOps[ss.Primary]
-	if err := pdb.SetReadOnly(ctx, true); err != nil {
-		return fmt.Errorf("failed to make instance %d read-only: %w", ss.Primary, err)
-	}
-	time.Sleep(100 * time.Millisecond)
-	if err := pdb.KillConnections(ctx); err != nil {
-		return fmt.Errorf("failed to kill connections in instance %d: %w", ss.Primary, err)
+
+	// SetReadOnly waits for a running DML.
+	// Therefore, if it waits for a long time, deleteGracePeriodSeconds may be reached.
+	// To avoid this, execute killConnections after a certain period of time.
+	done := make(chan error)
+	go func() {
+		done <- pdb.SetReadOnly(ctx, true)
+	}()
+	select {
+	case err := <-done:
+		if err != nil {
+			return fmt.Errorf("failed to make instance %d read-only: %w", ss.Primary, err)
+		}
+		time.Sleep(100 * time.Millisecond)
+		if err := pdb.KillConnections(ctx); err != nil {
+			return fmt.Errorf("failed to kill connections in instance %d: %w", ss.Primary, err)
+		}
+	case <-time.After(15 * time.Second):
+		log.Info("setReadOnly is taking too long, kill connections", "instance", ss.Primary)
+		if err := pdb.KillConnections(ctx); err != nil {
+			return fmt.Errorf("failed to kill connections in instance %d: %w", ss.Primary, err)
+		}
+		err := <-done
+		if err != nil {
+			return fmt.Errorf("failed to make instance %d read-only: %w", ss.Primary, err)
+		}
+		time.Sleep(100 * time.Millisecond)
+		if err := pdb.KillConnections(ctx); err != nil {
+			return fmt.Errorf("failed to kill connections in instance %d: %w", ss.Primary, err)
+		}
 	}
 	pst, err := pdb.GetStatus(ctx)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


clustering/operations.go
```diff
@@ -149,14 +149,36 @@ func (p *managerProcess) clone(ctx context.Context, ss *StatusSet) (bool, error)
...
-	if err := pdb.SetReadOnly(ctx, true); err != nil {
-		return fmt.Errorf("failed to make instance %d read-only: %w", ss.Primary, err)
+
+	// SetReadOnly waits for a running DML.
+	// Therefore, if it waits for a long time, deleteGracePeriodSeconds may be reached.
+	// To avoid this, set a timeout on SetReadOnly.
+	// If SetReadOnly fails, kill all processes and retry.
+	readOnly := false
+	for i := 0; i < 2; i++ {
+		ctxTimeout, cancel := context.WithTimeout(ctx, 15*time.Second)
```
ymmt2005, line 160: @shunki-fujita 
Can we use `SET SESSION wait_timeout` instead?
https://dev.mysql.com/doc/refman/8.4/en/server-system-variables.html#sysvar_wait_timeout

shunki-fujita: @ymmt2005 
Using `lock_wait_timeout` sounds like a good idea.
Thanks.

```
mysql> SET GLOBAL super_read_only=1;
ERROR 1205 (HY000): Lock wait timeout exceeded; try restarting transaction
mysql>
```

clustering/operations.go
```diff
@@ -151,12 +151,36 @@ func (p *managerProcess) switchover(ctx context.Context, ss *StatusSet) error {
...
-	if err := pdb.KillConnections(ctx); err != nil {
-		return fmt.Errorf("failed to kill connections in instance %d: %w", ss.Primary, err)
+
+	// SetReadOnly waits for a running DML.
+	// Therefore, if it waits for a long time, deleteGracePeriodSeconds may be reached.
+	// To avoid this, execute killConnections after a certain period of time.
+	done := make(chan error)
+	go func() {
+		done <- pdb.SetReadOnly(ctx, true)
+	}()
```
ymmt2005, line 161: This goroutine leaks if the control reaches line 169 because
the `done` channel is unbuffered.

I'd suggest making the channel buffered.
```suggestion
	done := make(chan error, 1)
	go func() {
		done <- pdb.SetReadOnly(ctx, true)
	}()
```

clustering/operations.go
```diff
@@ -151,12 +151,36 @@ func (p *managerProcess) switchover(ctx context.Context, ss *StatusSet) error {
...
+	select {
+	case err := <-done:
+		if err != nil {
+			return fmt.Errorf("failed to make instance %d read-only: %w", ss.Primary, err)
+		}
+		time.Sleep(100 * time.Millisecond)
+		if err := pdb.KillConnections(ctx); err != nil {
+			return fmt.Errorf("failed to kill connections in instance %d: %w", ss.Primary, err)
+		}
+	case <-time.After(15 * time.Second):
```
ymmt2005, line 171: Question.
How did you determine this 15-sec duration? Is there any reason?

shunki-fujita: @ymmt2005 
I thought the shorter the time it takes to switchover, the better, but I set it excessively short.
Waiting a little does not affect availability, so I'll wait as long as the WaitForGTID wait time.
`switchOverTimeoutSeconds = 70`

ymmt2005 (dismissed):
LGTM

clustering/operations.go
```diff
@@ -151,12 +151,36 @@ func (p *managerProcess) switchover(ctx context.Context, ss *StatusSet) error {
...
+	select {
+	case err := <-done:
+		if err != nil {
+			return fmt.Errorf("failed to make instance %d read-only: %w", ss.Primary, err)
+		}
+		time.Sleep(100 * time.Millisecond)
+		if err := pdb.KillConnections(ctx); err != nil {
+			return fmt.Errorf("failed to kill connections in instance %d: %w", ss.Primary, err)
+		}
+	case <-time.After(switchOverTimeoutSeconds):
```
ymmt2005, line 171:
```suggestion
	case <-time.After(switchOverTimeoutSeconds * time.Second):
```

ymmt2005 (approved):

clustering/operations.go
```diff
@@ -151,12 +151,36 @@ func (p *managerProcess) switchover(ctx context.Context, ss *StatusSet) error {
...
+	// Therefore, if it waits for a long time, deleteGracePeriodSeconds may be reached.
+	// To avoid this, execute killConnections after a certain period of time.
+	done := make(chan error, 1)
+	go func() {
+		done <- pdb.SetReadOnly(ctx, true)
+	}()
+	select {
+	case err := <-done:
+		if err != nil {
+			return fmt.Errorf("failed to make instance %d read-only: %w", ss.Primary, err)
```
shunki-fujita, line 165: [bug] If `switchOverTimeoutSeconds` is shorter than `lock_wait_timeout`, it returns here.

shunki-fujita:
The problem is that the switching is not completed within the PreStopSeconds time.
Solution will be considered separately.

Issue #725, opened 2024-08-06
Load only the named schema from the dump files


shunki-fujita:
## What
Restore only the specified schema from the dump file.

## How
Specify includeSchemas in loadDump options.
https://dev.mysql.com/doc/mysql-shell/8.0/en/mysql-shell-utilities-load-dump.html#mysql-shell-utilities-load-dump-opt-filtering

Specify schemas in RestoreSpec and pass it to LoadDump.
https://cybozu-go.github.io/moco/crd_mysqlcluster_v1beta2.html#restorespec
https://github.com/cybozu-go/moco/blob/f3c730d924febd491a36d39af039263df5ca352a/pkg/bkop/restore.go#L42

## Checklist

- [ ] Finish implementation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

PR #726, opened 2024-08-06
Load only the named schema from the dump files

shunki-fujita:
https://github.com/cybozu-go/moco/issues/725

### DIFF
--- a/api/v1beta2/mysqlcluster_types.go
+++ b/api/v1beta2/mysqlcluster_types.go
@@ -566,6 +566,10 @@ type RestoreSpec struct {
 
 	// Specifies parameters for restore Pod.
 	JobConfig `json:"jobConfig"`
+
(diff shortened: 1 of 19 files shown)


api/v1beta2/mysqlcluster_types.go
```diff
@@ -566,6 +566,10 @@ type RestoreSpec struct {
 
 	// Specifies parameters for restore Pod.
 	JobConfig `json:"jobConfig"`
+
+	// Schema is the name of the schema to restore.
+	// If empty, all schemas are restored.
+	Schema string `json:"schema,omitempty"`
```
yoheinbb, line 572: I think it would be good to add an +optional marker for kubebuilder.

yoheinbb (approved):

yoheinbb:
LGTM

PR #727, opened 2024-08-13
support k8s 1.30 and 1.31

YZ775:

PR #736, opened 2024-09-04
Update upload/download-artifact to v4

masa213f:


masa213f:
upload/download-artifact will be updated in https://github.com/cybozu-go/moco/pull/727

PR #737, opened 2024-09-06
Add a line break to kubectl-moco start/stop command's message

masa213f:


ymmt2005 (commented):
Why don't you just use `fmt.Fprintln`?

Since all of these output to `os.Stdout`, we can even use `fmt.Println`.

masa213f:
@ymmt2005 
That's true. fixed.

ymmt2005 (approved):
LGTM

PR #816, opened 2025-07-09
exclude replicas with uncommitted trx during failover

### LINKED ISSUE #813: The failover process cannot be recovered.
## What

If the primary pod has uncommitted transactions at the time failover processing begins, `configureReplica()` will fail.

In moco's failover processing, `failover()` first stops the IO_THREAD of the replica, then determines which pod will become the primary.
Then, `configure()` actually switches the primary and replica.
At this point, if the old primary is attempting to commit an uncommitted transaction, it cannot commit because the replica's IO_THREAD has already been stopped.
The connections existing at that time are disconnected via `KillConnection()`, but if the transaction is still waiting to commit, this step is passed normally.
```
mysql> show full processlist;
+------+-----------------+------------------+------+---------+-------+--------------------------------------+------------------------------+
| Id   | User            | Host             | db   | Command | Time  | State                                | Info                         |
+------+-----------------+------------------+------+---------+-------+--------------------------------------+------------------------------+
|    5 | event_scheduler | localhost        | NULL | Daemon  | 71234 | Waiting on empty queue               | NULL                         |
| 7852 | moco-writable   | localhost        | test | Killed  |    94 | Waiting for semi-sync ACK from slave | commit                       |
| 7853 | moco-admin      | localhost        | NULL | Query   |    36 | Waiting for commit lock              | set global super_read_only=1 |
| 7854 | moco-writable   | localhost        | test | Sleep   |    68 |                                      | NULL                         |
| 7855 | moco-agent      | 10.244.3.6:34992 | NULL | Sleep   |     2 |                                      | NULL                         |
| 7856 | moco-admin      | localhost        | NULL | Query   |     0 | init                                 | show full processlist        |
+------+-----------------+------------------+------+---------+-------+--------------------------------------+------------------------------+
6 rows in set, 1 warning (0.00 sec)
```

Therefore, the `set super_read_only=1` performed in `configureReplica()` fails.

Once this state is reached, the failover process continues to fail, and this cannot be automatically recovered.

## How

If a Pod is a replica and not a candidate, and has `rpl_semi_sync_master_wait_sessions > 0`, mark that Pod as failing.

## Checklist

- [x] Finish implementation of the issue
- [x] Test all functions
- [x] Have enough logs to trace activities
- [x] Notify developers of necessary actions

lrf141:
Closed #813

### DIFF
--- a/clustering/status.go
+++ b/clustering/status.go
@@ -216,10 +216,10 @@ func (p *managerProcess) GatherStatus(ctx context.Context) (*StatusSet, error) {
 				}
 				// process errors
 				if j == statusCheckRetryMax {
-					logFromContext(ctx).Error(err, "failed to get mysqld status, mysqld is not ready", "instance", index)
+					log.Error(err, "failed to get mysqld status, mysqld is not ready", "instance", index)
 					return
 				}
-				logFromContext(ctx).Error(err, "failed to get mysqld status, will retry", "instance", index)
+				log.Error(err, "failed to get mysqld status, will retry", "instance", index)
 				time.Sleep(statusCheckRetryInterval)
 			}
 		}(i)
--- a/clustering/status.go
+++ b/clustering/status.go
@@ -140,6 +140,7 @@ func (ss *StatusSet) DecideState() {
 // StatusSet.  It calls `StatusSet.DecideState` before returning.
 func (p *managerProcess) GatherStatus(ctx context.Context) (*StatusSet, error) {
 	ss := &StatusSet{}
+	log := logFromContext(ctx)
 
 	cluster := &mocov1beta2.MySQLCluster{}
 	if err := p.reader.Get(ctx, p.name, cluster); err != nil {
@@ -298,6 +299,7 @@ func (p *managerProcess) GatherStatus(ctx context.Context) (*StatusSet, error) {
 		}
 		if ist.GlobalStatus.SemiSyncMasterWaitSessions > 0 {
 			ss.MySQLStatus[i] = nil
+			log.Info("Detected a hangup replica. rpl_semi_sync_master_wait_sessions is greater than 0.", "instance", i)
 		}
 	}
 
--- a/clustering/status.go
+++ b/clustering/status.go
@@ -210,11 +210,7 @@ func (p *managerProcess) GatherStatus(ctx context.Context) (*StatusSet, error) {
 					return
 				}
 				if err == nil {
-					if index != ss.Primary && index != ss.Candidate && ist.GlobalStatus.SemiSyncMasterWaitSessions > 0 {
-						ss.MySQLStatus[index] = nil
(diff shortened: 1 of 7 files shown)


clustering/status.go
```diff
@@ -210,7 +210,11 @@ func (p *managerProcess) GatherStatus(ctx context.Context) (*StatusSet, error) {
 					return
 				}
 				if err == nil {
-					ss.MySQLStatus[index] = ist
+					if index != ss.Primary && index != ss.Candidate && ist.GlobalStatus.SemiSyncMasterWaitSessions > 0 {
```
yamatcha: ss.Candidate check seems unnecessary here since where it's assigned later.
https://github.com/cybozu-go/moco/blob/c9cf46426e319ee19be4a4cc7ee58176dab392e6/clustering/operations.go#L239
https://github.com/cybozu-go/moco/blob/c9cf46426e319ee19be4a4cc7ee58176dab392e6/clustering/status.go#L135

pkg/dbop/status.go
```diff
@@ -48,6 +55,36 @@ func (o *operator) getGlobalVariablesStatus(ctx context.Context) (*GlobalVariabl
 	return status, nil
 }
 
+func (o *operator) getGlobalStatus(ctx context.Context) (*GlobalStatus, error) {
+	status := &GlobalStatus{}
+	rows, err := o.db.QueryxContext(ctx, "SHOW GLOBAL STATUS LIKE '%Rpl_semi%'")
```
arosh: * `SELECT ... FROM performance_schema.global_status WHERE VARIABLE_NAME = 'Rpl_semi_sync_master_wait_sessions'` is preferred.
* Use [sqlx's GetContext](https://jmoiron.github.io/sqlx/#getAndSelect) like other functions.

clustering/status.go
arosh: Moreover, the purpose of this `go func()` is to gather MySQL status. The logic based on the gathered results should be implemented after "re-check the primary MySQL status to retrieve the latest executed GTID set". This is similar to "detect replication delay" or "detect errant replicas".

pkg/dbop/status_test.go
```diff
@@ -2,6 +2,7 @@ package dbop
 
 import (
 	"context"
+	"github.com/cybozu-go/moco/pkg/constants"
 
 	mocov1beta2 "github.com/cybozu-go/moco/api/v1beta2"
```
ymmt2005: The format does not look right.
```suggestion

	"github.com/cybozu-go/moco/pkg/constants"
	mocov1beta2 "github.com/cybozu-go/moco/api/v1beta2"
```

yoheinbb (approved):

yoheinbb (approved):

yamatcha (approved):

PR #819, opened 2025-07-21
Support for k8s v1.33

kdambekalns:


lrf141:
@kdambekalns 
Okay, now i approved the CI run.
We are also planning to support k8s v1.33, so we would be very happy if you could submit this pull request.

yamatcha:
@kdambekalns 
Thank you for your contribution!

The CI failure appears to be caused by an outdated version of kind and version incompatibility between kind and the kindest node image.

Could you please upgrade kind and specify KUBERNETES_VERSION with `1.33.1`?
You can update the kind version in this file: https://github.com/cybozu-go/moco/blob/efd1a3b735473572ee659f867b81e516536572fe/aqua.yaml#L14

yamatcha:
Sorry, I missed this.  It's also necessary to update the other references to 1.33.0 to maintain consistency.
for example: https://github.com/cybozu-go/moco/blob/db23835f08ce01e93231a5af8c4189d21da22393/.github/workflows/ci-e2e.yaml#L44

kdambekalns:
Ah, could have checked that myself… let's see what happens now.

kdambekalns:
Ah, (probably) very simple, after all…

.github/workflows/helm.yaml
```diff
@@ -21,7 +21,7 @@ jobs:
           python-version: 3.8
 
       - name: Set up chart-testing
-        uses: helm/chart-testing-action@0d28d3144d3a25ea2cc349d6e59901c4ff469b3b # v2.7.0
```
yamatcha: We specify these actions with hash values to ensure we use the correct and secure versions of the actions.
Please revert these changes to use the hash values instead of version tags.

.github/workflows/helm.yaml
```diff
@@ -34,12 +34,12 @@ jobs:
...
 
       - name: Create kind cluster
-        uses: helm/kind-action@a1b0e391336a6ee6713a0583f8c6240d70863de3 # v1.12.0
+        uses: helm/kind-action@v1.12.0
         if: steps.list-changed.outputs.changed == 'true'
         with:
-          version: v0.23.0
-          node_image: kindest/node:v1.32.0
-          kubectl_version: v1.32.0
+          version: v0.24.0
```
yamatcha: This kind version should match the version specified in aqua.yaml.

```suggestion
          version: v0.29.0
```

yamatcha (commented):
Just a few minor nitpicks in the comments. Sorry for the scattered feedback.

.github/workflows/helm.yaml
kdambekalns: Fine – but why here and not consistently everywhere or nowhere? Is there some background story I should know?

yamatcha: Thank you for the fix. Our policy applies mainly to third-party actions.
You don't need to worry about the background since those actions are already pinned to specific hash values.

yamatcha (approved):
LGTM!

lrf141 (approved):
I was scheduled to perform this task, but thanks to your contribution, we were able to provide support earlier than planned.
Thank you.

LGTM!

kdambekalns:
🎉

PR #825, opened 2025-08-07
Fix flaky test: wait for SemiSyncMasterWaitSessions to update after replica hang

yamatcha:
After merging #816, the `dbtest` frequently fails due to a flaky test.

The test expects `SemiSyncMasterWaitSessions` to become non-zero immediately after creating a hanging transaction, but there can be a timing issue where the value hasn't updated yet when the assertion runs.

This PR fixes the issue by using `Eventually()` to wait for the expected state instead of checking it immediately.

Failing CI: https://github.com/cybozu-go/moco/actions/runs/16764566851/job/47467091885

### DIFF
--- a/pkg/dbop/status_test.go
+++ b/pkg/dbop/status_test.go
@@ -129,7 +129,13 @@ var _ = Describe("status", func() {
 		st1, err = ops[1].GetStatus(ctx)
 		Expect(err).NotTo(HaveOccurred())
 		Expect(st0.GlobalVariables.ExecutedGTID).To(Equal(st1.GlobalVariables.ExecutedGTID))
-		Expect(st0.GlobalStatus.SemiSyncMasterWaitSessions).NotTo(Equal(0))
+		Eventually(func() int {
+			st0, err = ops[0].GetStatus(ctx)
+			if err != nil {
+				return 0
+			}
+			return st0.GlobalStatus.SemiSyncMasterWaitSessions
+		}).ShouldNot(Equal(0))
 		cancelTrx()
 	})
 })


pkg/dbop/status_test.go
```diff
@@ -129,7 +129,9 @@ var _ = Describe("status", func() {
 		st1, err = ops[1].GetStatus(ctx)
 		Expect(err).NotTo(HaveOccurred())
 		Expect(st0.GlobalVariables.ExecutedGTID).To(Equal(st1.GlobalVariables.ExecutedGTID))
-		Expect(st0.GlobalStatus.SemiSyncMasterWaitSessions).NotTo(Equal(0))
+		Eventually(func() {
```
yoheinbb: Isn't it necessary to update ops[0].GetStatus(ctx)?
And it seems better to use Eventually together with Should or ShouldNot.

```suggestion
        Eventually(func() int {
            st0, _ := ops[0].GetStatus(ctx)
            return st0.GlobalStatus.SemiSyncMasterWaitSessions
        }).ShouldNot(Equal(0))
```

pkg/dbop/status_test.go
```diff
@@ -129,7 +129,13 @@ var _ = Describe("status", func() {
 		st1, err = ops[1].GetStatus(ctx)
 		Expect(err).NotTo(HaveOccurred())
 		Expect(st0.GlobalVariables.ExecutedGTID).To(Equal(st1.GlobalVariables.ExecutedGTID))
-		Expect(st0.GlobalStatus.SemiSyncMasterWaitSessions).NotTo(Equal(0))
+		Eventually(func() (int, error) {
```
yoheinbb: It seems that err does not need to be included in the return value.
ref. https://github.com/cybozu-go/moco/blob/068150d7dd59619036dd364f3f798abd399845ce/pkg/dbop/status_test.go#L92-L99

yamatcha: https://github.com/cybozu-go/moco/pull/825/commits/5414f4b5ef6f91d31179dc0e2c05f37d3175c463

yoheinbb (approved):

PR #830, opened 2025-08-28
Long wait for SetReadOnly to complete

### LINKED ISSUE #826: Long wait for SetReadOnly to complete
## What

SET READ_ONLY waits for queries that use metadata locks.
(Including select xx from xx for update queries)
Therefore, if the wait time until READ_ONLY is set becomes long, deleteGracePeriodSeconds will be reached and a failover will occur.

This issue was previously canceled, but improvements can be expected by using a time shorter than PreStopSeconds.
#708

## How

If SET READ_ONLY does not complete in a short time, give up and run KillConnections.

## Checklist

- [ ] Test all functions

lrf141:
Close #826

### DIFF
--- a/e2e/switchover_test.go
+++ b/e2e/switchover_test.go
@@ -60,6 +60,11 @@ var _ = Context("switchover", Ordered, func() {
 	})
 
 	It("should switch the primary if requested, even when a long global read lock is acquired", func() {
+		cluster, err := getCluster("switchover", "test")
+		Expect(err).NotTo(HaveOccurred())
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,3 +1,5 @@
+KUBERNETES_VERSION = 1.33.0
```
lrf141, line 1: I would prefer to use 1.33.1, but since `setup-envtest` cannot find the 1.33.1 archive.

yoheinbb (approved):

Issue #834, opened 2025-09-12
Allow setting a custom user for sidecar containers


kevinrudde:
## What

Currently, all sidecars run with the default security context configuration, but some sidecars (e.g., Tailscale) require running as root to function properly.

## How

Default to the existing non-root user for safety.
Allow overriding to root or other users when explicitly set by the operator.
Be applied consistently across all sidecar containers without affecting the main MySQL containers.

I think currently the security context gets overridden here: https://github.com/cybozu-go/moco/blob/28c8a3242a0b3ba75b2a18349befd1bc789e43dd/controllers/mysql_container.go#L291C1-L291C41

## Checklist

- [ ] Finish implementation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

PR #835, opened 2025-09-15
fix: only set runAsUser and runAsGroup if they are not specified

kevinrudde:
See https://github.com/cybozu-go/moco/issues/834


shunki-fujita:
@kevinrudde
LGTM. Thanks!

PR #857, opened 2025-11-11
Correct namespace scoping of PVC resize targets

### LINKED ISSUE #854: Correct namespace scoping of PVC resize targets
## What

A logic review of pvc.go showed the PVC resize target discovery must be strictly limited to the owning MySQLCluster’s namespace. Without explicit scoping, PVCs in other namespaces that share the StatefulSet’s labels could be (or could previously have been) included unintentionally.

## How

List PVCs with client.InNamespace(sts.Namespace) and the StatefulSet’s selector

## Checklist

- [ ] Test all functions

yamatcha:
closed #854

### DIFF
--- a/controllers/partition_controller.go
+++ b/controllers/partition_controller.go
@@ -15,6 +15,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/kubectl/pkg/util/podutils"
+	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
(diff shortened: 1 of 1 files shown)


controllers/partition_controller.go
```diff
@@ -138,7 +138,7 @@ func (r *StatefulSetPartitionReconciler) SetupWithManager(mgr ctrl.Manager) erro
 func (r *StatefulSetPartitionReconciler) isRolloutReady(ctx context.Context, cluster *mocov1beta2.MySQLCluster, sts *appsv1.StatefulSet) (bool, error) {
 	log := crlog.FromContext(ctx)
 
-	if sts.Spec.Replicas == &sts.Status.UpdatedReplicas {
+	if *sts.Spec.Replicas == sts.Status.UpdatedReplicas {
```
arosh: ptr.Deref is preferred to avoid panic.

PR #860, opened 2025-11-27
Support k8s 1.34

### LINKED ISSUE #841: support k8s 1.34
## What

support k8s 1.34

## Checklist

- [ ] Test all functions

lrf141:
closed #841

### DIFF
--- a/.github/actions/dbtest/action.yaml
+++ b/.github/actions/dbtest/action.yaml
@@ -9,7 +9,7 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - uses: actions/setup-go@v5
+    - uses: actions/setup-go@v6
       with:
         go-version-file: go.mod
         cache: true
--- a/.github/actions/e2e/action.yaml
+++ b/.github/actions/e2e/action.yaml
@@ -14,7 +14,7 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - uses: actions/setup-go@v5
+    - uses: actions/setup-go@v6
       with:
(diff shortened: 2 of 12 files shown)


controllers/mysqlcluster_controller.go
```diff
@@ -126,11 +126,11 @@ func apply[S any, T clientObjectConstraint[S], U any](ctx context.Context, r cli
...
 		return nil, fmt.Errorf("failed to convert resource %s/%s to unstructured: %w", key.Namespace, key.Name, err)
 	}
-	patch := &unstructured.Unstructured{
+	patch := client.ApplyConfigurationFromUnstructured(&unstructured.Unstructured{
 		Object: obj,
-	}
+	})
 
-	return &orig, r.Patch(ctx, patch, client.Apply, &client.PatchOptions{
+	return &orig, r.Apply(ctx, patch, &client.ApplyOptions{
```
lrf141, line 133: `client.Apply` is deprecated and will fail static checks.
It is now recommended to use `r.Apply` instead of `r.Patch`.
As a result, `unstructured.Unstructured` can no longer be used as an argument, but I decided to convert it to `ApplyConfiguration` this time.

This implementation is somewhat unclear, so I think generating `ApplyConfiguration` might be better in the future. However, due to the large scope of impact, I will not do this at this time.

lrf141: At least with this implementation, the tests pass.

yamatcha (approved):

Issue #867, opened 2026-01-19
Allow restoring specific users.


yoheinbb:
## What

Allow restoring only specified users.

## How

Implement the includeUsers option to allow for the restoration of only specified users.

Specify includeUsers  in loadDump options.
https://dev.mysql.com/doc/mysql-shell/8.4/en/mysql-shell-utilities-load-dump.html#mysql-shell-utilities-load-dump-opt-filtering

Specify users in RestoreSpec and pass it to LoadDump.
https://cybozu-go.github.io/moco/crd_mysqlcluster_v1beta2.html#restorespec

https://github.com/cybozu-go/moco/blob/4466e474b08ca299586dac5a5430d498beadd890/pkg/bkop/restore.go#L45

## Checklist

- [x] Finish implementation of the issue
- [x] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

PR #872, opened 2026-01-21
Add Users to RestoreSpec

yoheinbb:
https://github.com/cybozu-go/moco/issues/867

### DIFF
--- a/controllers/mysqlcluster_controller_test.go
+++ b/controllers/mysqlcluster_controller_test.go
@@ -1583,6 +1583,8 @@ dummyKey: dummyValue
 			SourceName:      "single",
 			SourceNamespace: "ns",
 			RestorePoint:    now,
+			Schema:          "db1",
+			Users:           "user1,user2@%,user3@localhost",
 		}
 		jc := &cluster.Spec.Restore.JobConfig
 		jc.Threads = 3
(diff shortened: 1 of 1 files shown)


controllers/mysqlcluster_controller_test.go
```diff
@@ -1672,6 +1672,7 @@ dummyKey: dummyValue
 			"test",
 			now.UTC().Format(constants.BackupTimeFormat),
 			"",
+			"",
```
yamatcha: Please confirm the controller construct these arguments using `cluster.Spec.Restore.Schema` and `cluster.Spec.Restore.Users`.

e2e/backup_test.go
```diff
@@ -181,16 +191,24 @@ var _ = Context("backup", Ordered, func() {
...
 		Expect(err).NotTo(HaveOccurred())
 		Expect(count).To(Equal(2))
 
+		// Verify that only test2 schema exists
 		out = kubectlSafe(nil, "moco", "-n", "backup", "mysql", "target2", "--",
 			"-N", "-e", "SHOW DATABASES LIKE 'test%'")
 		databases := strings.Fields(string(out))
 		Expect(databases).Should(ConsistOf("test2"))
+
+		// Verify that only test_user2 exists
```
yamatcha, line 207: This check only verifies that `test_user2` exists, but it does not ensure that it is the only user. 
Please add an additional check to confirm that `test_user1` does not exist.

yoheinbb: This check retrieves test_user% using the following query and verifies that the result is only test_user2.
```
SELECT User FROM mysql.user WHERE User LIKE 'test_user%'
```

yamatcha: Sorry, I mistook `ConsistOf` for `Contains`...

controllers/mysqlcluster_controller_test.go
yoheinbb: I fix it
2a8687147a22341182a284a7022257b95dd865c7

PR #880, opened 2026-02-06
Refactor owner reference helpers with generics

arosh:
## Summary  

Introduce a generic helper to set OwnerReferences on ApplyConfiguration objects. Replace per-resource owner reference helpers.

## Background  

The controller used multiple near-identical functions to set OwnerReferences on different ApplyConfiguration types, making maintenance noisy and repetitive.

## Changes  

- Added a generic `setControllerReference` helper for ApplyConfiguration objects.  
- Updated all non-PVC owner reference calls to use the generic helper.  
- Removed redundant per-resource helper functions, retaining the PVC-specific one.

## Note

Keep the PVC-specific helper unchanged for its APIVersion handling.

### DIFF
--- a/controllers/mysqlcluster_controller.go
+++ b/controllers/mysqlcluster_controller.go
@@ -372,7 +372,7 @@ func (r *MySQLClusterReconciler) reconcileUserSecret(ctx context.Context, req ct
 		WithLabels(labelSet(cluster, false)).
 		WithData(newSecret.Data)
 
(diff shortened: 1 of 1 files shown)


lrf141 (approved):

lrf141 (commented):
@arosh 
Signatures are required for commits.
https://github.com/cybozu-go/moco/blob/main/README.md#contributing

PR #887, opened 2026-02-18
Add imagePullSecrets support to JobConfig for backup and restore jobs

### LINKED ISSUE #848: Add support for imagePullSecrets for Backup and Restore
Hi! We use private container registry with authorization.

CR MySQLCluster support imagePullSecrets in the spec.podTemplate.spec, but not in spec.restore.jobConfig.
Also CR BackupPolicy doesn't support it.

Fortunately we can patch created by Operator backup CronJob resource manually, and Operator doesn't prevent imagePullSecrets addition.
But we can't patch created by Operator restore job.batch, so before restoration we have to pull a necessary image to nodes manually.

That would be great to add imagePullSecrets optional specification for BackupPolicy and inherit imagePullSecrets for restore job from MySQLCluster CR.

shunki-fujita:
Close: #848 

Users with private container registries can now specify imagePullSecrets in JobConfig, which is propagated to both backup CronJob and restore Job PodSpecs. This enables pulling backup/restore images from authenticated registries without manual intervention.


api/v1beta2/job_types.go
```diff
@@ -86,6 +86,11 @@ type JobConfig struct {
 	//
 	// +optional
 	VolumeMounts []VolumeMountApplyConfiguration `json:"volumeMounts,omitempty"`
+
+	// ImagePullSecrets is a list of references to secrets in the same namespace to use for pulling any images.
```
arosh: Please clarify in the comment how this `imagePullSecrets` in JobConfig differs from the `imagePullSecrets` in `podTemplate.spec`. 

For example:

> "ImagePullSecrets is a list of references to secrets in the same namespace to use for pulling backup and restore job images. This applies specifically to the backup CronJob and restore Job Pods. To configure imagePullSecrets for the MySQLCluster's StatefulSet Pods, use `podTemplate.spec.imagePullSecrets` instead.

Since this description will also be used in the CRD documentation (docs/crd_mysqlcluster_v1beta2.md), it's important to make the distinction clear for users who might wonder about the relationship between these two fields.

arosh (dismissed):

controllers/mysqlcluster_controller_test.go
```diff
@@ -112,6 +112,9 @@ func testNewBackUpPolicy() *mocov1beta2.BackupPolicy {
 			MountPath: ptr.To[string]("/path/to/dir"),
 		},
 	}
+	jc.ImagePullSecrets = []mocov1beta2.LocalObjectReferenceApplyConfiguration{
+		{Name: ptr.To[string]("my-registry-secret")},
```
yoheinbb, line 116: @shunki-fujita 
Please update your branch with the latest changes from main.
Also, please replace the k8s.io/utils/ptr.To() calls with Go 1.26's new new(expr) builtin once you've updated. That would let us drop the k8s.io/utils/ptr dependency here.

PR #903, opened 2026-03-19
System User Password Rotation with CredentialRotation CRD

### LINKED ISSUE #849: System users password management
It would be great to add support for system users password management.

When system users passwords are compromised, it's be impossible to change them in an existing cluster.

I tried to deploy Secret with passwords for users:
- moco-admin
- moco-agent
- moco-backup
- moco-clone-donor
- moco-exporter
- moco-readonly
- moco-repl
- moco-writable
And manually change their passwords in MySQL, but the operator rolled back my changes.

So, it seems, the only way to change system users passwords is to deploy a new cluster and move all data to it.

Maybe I'm wrong, and there's an easier way?

shunki-fujita:
Close: https://github.com/cybozu-go/moco/issues/849

Propose a dedicated CredentialRotation CRD with its own controller as an alternative to embedding password rotation into MySQLCluster.


pkg/dbop/password.go
```diff
@@ -0,0 +1,220 @@
...
+// relevant for MOCO system users.
+//
+// If the first element is '*' (any plugin allowed) or empty, "caching_sha2_password"
+// is returned as the default.
+func (o *operator) GetAuthPlugin(ctx context.Context) (string, error) {
+	var policy string
+	if err := o.db.GetContext(ctx, &policy, "SELECT @@global.authentication_policy"); err != nil {
+		return "", fmt.Errorf("failed to get authentication_policy: %w", err)
+	}
+	return parseAuthPlugin(policy)
```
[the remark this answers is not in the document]
shunki-fujita: moco supports MySQL 8.0.28 and later

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,776 @@
...
+### Why Auth Plugin Migration Is in the Discard Phase?
+
+MySQL Error 3894 prevents changing the authentication plugin in a `RETAIN CURRENT PASSWORD` statement.
+Instead, plugin migration happens after DISCARD using `ALTER USER ... IDENTIFIED WITH <plugin> BY ...`.
+
+The target plugin is determined from `@@global.authentication_policy` on the primary instance:
+- If the first element is a concrete plugin name (e.g. `caching_sha2_password`): use it
+- If the first element is `*` or empty: default to `caching_sha2_password`
+
+This enables transparent migration from legacy plugins like `mysql_native_password` during rotation.
```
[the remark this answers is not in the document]
shunki-fujita: moco supports MySQL 8.0.28 and later

cmd/kubectl-moco/cmd/credential.go
```diff
@@ -46,7 +100,103 @@ password="%s"
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      clusterName,
+				Namespace: namespace,
+			},
+			Spec: mocov1beta2.CredentialRotationSpec{
+				RotationGeneration: 1,
+				DiscardOldPassword: false,
+			},
+		}
+		// ownerReference is set by the controller on first reconcile.
```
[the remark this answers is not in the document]
shunki-fujita: The race window between CR creation and the first controller reconcile is negligibly small, and an orphaned CR is harmless (it does nothing without a matching MySQLCluster). Keeping ownerReference setup solely in the controller avoids duplicating the logic.

clustering/password_rotation.go
```diff
@@ -0,0 +1,368 @@
...
+			if checkErr != nil {
+				return false, checkErr
+			}
+			if dualFound {
+				log.Info("waiting: instance has pre-existing dual password",
+					"instance", idx, "user", dualUser)
+				p.recorder.Eventf(cluster, corev1.EventTypeWarning, "DualPasswordExists",
+					"Cannot proceed with RETAIN: instance %d user %s already has a dual password. "+
+						"See MOCO documentation for recovery procedures.",
+					idx, dualUser)
```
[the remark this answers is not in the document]
shunki-fujita: Kubernetes EventRecorder aggregates identical events automatically (incrementing count instead of creating duplicates), so the practical impact is minimal. The proposed annotation-based deduplication adds complexity and an extra Secret update per loop iteration, which outweighs the benefit.

controllers/credentialrotation_controller.go
```diff
@@ -0,0 +1,397 @@
...
+}
+
+// handleStartRotation: ""/Completed → Rotating
+func (r *CredentialRotationReconciler) handleStartRotation(ctx context.Context, cr *mocov1beta2.CredentialRotation, cluster *mocov1beta2.MySQLCluster) (ctrl.Result, error) {
+	log := crlog.FromContext(ctx)
+
+	if cluster.Spec.Replicas <= 0 {
+		r.Recorder.Eventf(cr, corev1.EventTypeWarning, "RotationRefused",
+			"Cannot start rotation: MySQLCluster replicas is 0")
+		return ctrl.Result{RequeueAfter: credRotationRequeueInterval}, nil
```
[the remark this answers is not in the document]
shunki-fujita: Same as above — EventRecorder aggregates repeated events. Also, removing the requeue would cause the reconciler to stall after scale-up, since CredentialRotationReconciler does not watch MySQLCluster resources.

clustering/bin/1.34.0-linux-amd64/etcd
arosh, line 1: DO NOT push binary files

arosh (changes requested):

.gitignore
```diff
@@ -19,4 +19,5 @@
 /bin
 /build
 /testbin
+clustering/bin/
```
arosh: Why is this line required?

shunki-fujita: `clustering/bin/` only appears when running tests outside `make clustering-envtest` (e.g. IDE or bare `go test ./clustering` with no `ENVTEST_ASSETS_DIR`).
The standard flow writes to `/bin`, which is already ignored.
It isn't needed and I'm removing it.

api/v1beta2/credentialrotation_types.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+// CredentialRotationSpec defines the desired state of CredentialRotation.
+// The target MySQLCluster is identified by the CR's own name and namespace
+// (CredentialRotation name must equal MySQLCluster name).
+type CredentialRotationSpec struct {
+	// RotationGeneration is a monotonically increasing counter.
+	// Incrementing this value triggers a new rotation cycle.
+	// +optional
+	RotationGeneration int64 `json:"rotationGeneration,omitempty"`
```
arosh: Please read the Kubernetes API convention before creating a new CustomResource.
https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#serialization-of-optionalrequired-fields

`RotationGeneration=0` is a valid value, right? How do you distinguish between a valid `RotationGeneration=0` and an omitted expression?

arosh (changes requested):

api/v1beta2/credentialrotation_types.go
```diff
@@ -0,0 +1,88 @@
...
+// +kubebuilder:printcolumn:name="ObservedDiscard",type="integer",JSONPath=".status.observedDiscardGeneration"
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
+
+// CredentialRotation is the Schema for the credentialrotations API
+type CredentialRotation struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   CredentialRotationSpec   `json:"spec,omitempty"`
+	Status CredentialRotationStatus `json:"status,omitempty"`
```
arosh: FIELDS MUST BE EITHER OPTIONAL OR REQUIRED!!
https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#serialization-of-optionalrequired-fields:~:text=Fields%20must%20be%20either%20optional%20or%20required.

arosh:
@shunki-fujita 
Why don't you ask to Claude Code to adhere to Kubernetes API Convention? Please the following into CLAUDE.md.

```text
When you edit any file matching the glob `api/**/*_types.go`, you MUST comply with the
[Kubernetes API Convention](https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md).
```

api/v1beta2/credentialrotation_types.go
```diff
@@ -0,0 +1,95 @@
...
+
+	// DiscardGeneration is a monotonically increasing counter that triggers
+	// the discard phase. Must satisfy 0 <= discardGeneration <= rotationGeneration.
+	// Bumping this value (typically to match rotationGeneration) signals the
+	// controller to discard the retained old password from the previous
+	// rotation. The bump is only honored when Phase is Rotated.
+	// +kubebuilder:default=0
+	// +kubebuilder:validation:Minimum=0
+	// +optional
+	DiscardGeneration int64 `json:"discardGeneration,omitempty"`
```
arosh: How do you distinguish between a valid `DiscardGeneration=0` and an omitted `PATCH` expression?

api/v1beta2/credentialrotation_types.go
```diff
@@ -0,0 +1,95 @@
...
+
+// CredentialRotationStatus defines the observed state of CredentialRotation.
+type CredentialRotationStatus struct {
+	// Phase is the current rotation phase.
+	// +optional
+	Phase RotationPhase `json:"phase,omitempty"`
+
+	// RotationID is the UUID for this rotation cycle.
+	// +optional
+	RotationID string `json:"rotationID,omitempty"`
```
arosh: Add validation or be a pointer

api/v1beta2/credentialrotation_types.go
```diff
@@ -0,0 +1,95 @@
...
+
+	// ObservedRotationGeneration is the last rotationGeneration
+	// that completed successfully.
+	// +kubebuilder:validation:Minimum=0
+	// +optional
+	ObservedRotationGeneration int64 `json:"observedRotationGeneration,omitempty"`
```
arosh: pointer?

api/v1beta2/credentialrotation_types.go
```diff
@@ -0,0 +1,95 @@
...
+// +kubebuilder:printcolumn:name="ObservedDiscard",type="integer",JSONPath=".status.observedDiscardGeneration"
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
+
+// CredentialRotation is the Schema for the credentialrotations API
+type CredentialRotation struct {
+	metav1.TypeMeta `json:",inline"`
+	// +optional
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec CredentialRotationSpec `json:"spec"`
```
arosh: required or optional?

yamatcha:
@shunki-fujita 
The current Status.Conditions shape violates the [Kubernetes API conventions](https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties). Conditions are supposed to be used as follows:

- Type names are "an adjective ('Ready', 'OutOfDisk') or a past-tense verb ('Succeeded', 'Failed') rather than a present-tense verb ('Rotating ')"
- "conditions are observations and not, themselves, state machines" — workflow steps don't belong in conditions
- Each condition is one of the "individual conditions that clients need to monitor" — i.e. orthogonal, independent observations
- Reason is "a programmatic identifier" — the same identifier should not mean different things in different conditions

### Proposal: replace with truly independent observations
For example: RotationReady, DiscardReady, OldPasswordRetained, etc...

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,629 @@
...
+```
+
+## Impact Summary
+
+| Category | Files |
+|---|---|
+| **New** | `api/v1beta2/credentialrotation_types.go`, `api/v1beta2/credentialrotation_helpers.go`, `api/v1beta2/credentialrotation_webhook.go`, `controllers/credentialrotation_controller.go`, `clustering/password_rotation.go` |
+| **New (CLI)** | `cmd/kubectl-moco/cmd/credential.go` (`rotate` / `discard` / `show` subcommands, stale-CR detection) |
+| **New (DB ops)** | `pkg/password/rotation.go`, `pkg/dbop/password.go` |
+| **Modified** | `controllers/mysqlcluster_controller.go` (`reconcileV1Secret` chooses current vs pending password by `cr.Step()` and self-heals per-namespace Secrets), `cmd/moco-controller/cmd/run.go` (register the new reconciler) |
```
arosh: meaningless

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,629 @@
...
+
+- `rotation` when `spec.rotationGeneration > status.observedRotationGeneration`
+- `discard` when `spec.discardGeneration > status.observedDiscardGeneration`
+- `—` when both are in sync
+
+The two values are mutually exclusive (the rotation phase always completes — promoting `observedRotationGeneration` — before the operator can request discard), which is what lets the matrix disambiguate the three `(False, False, True)` rows: `DistributingPassword` has `rotation` outstanding, `AwaitingRollout` has neither outstanding, and `ApplyingDiscard` has `discard` outstanding.
+
+A `Status=False` Reason of `Refused`/`Blocked`/`Stale` takes priority: it short-circuits to `RotationRefused` / `RotationBlocked` / `DiscardRefused` / `DiscardBlocked` / `StalePending` regardless of the table above. A stale `RotationReady=True` lingering across a fresh `rotationGeneration` bump is treated as Idle (and similarly for `DiscardReady=True` and AwaitingDiscard) so the seed handler fires — this avoids the "stale True from previous cycle" deadlock for back-to-back rotations.
```
arosh:
```suggestion
A `Status=False` condition with the `Refused`, `Blocked`, or `Stale`
reason takes precedence over the normal Step matrix. The corresponding
refused, blocked, or stale Step is selected immediately.

When a new `rotationGeneration` is requested, an old
`RotationReady=True` condition may remain until the next status update.
This old condition is treated as the previous cycle's state, so the CR is
still handled as Idle and the rotation seed handler can start the new cycle.
The same rule applies to an old `DiscardReady=True` condition and
`AwaitingDiscard`. This prevents a stale condition from the previous cycle
from stopping a subsequent rotation or discard.
```

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,629 @@
...
+
+#### Why three conditions?
+
+`RotationReady` and `DiscardReady` are action-availability guards — analogous to Pod `Ready=True` meaning "you may route traffic to me now". The two Ready conditions are structurally mutually exclusive (the Idle and AwaitingDiscard steady states cannot both hold), which removes the read-time ambiguity earlier "generation tracking" semantics had. `DualPassword` is the orthogonal physical-state observation, named as a noun in parallel with `MemoryPressure`-style conditions where `True` describes the situation rather than health.
+
+The current step is derived from these three conditions plus the generation comparisons, so the lack of a stored Phase keeps the CR convention-compliant without requiring extra API I/O at runtime.
```
arosh:
```suggestion
`RotationReady` and `DiscardReady` show whether the operator may start the next action. They are similar to a Pod's `Ready=True` condition, which means that the Pod can receive traffic. These two Ready conditions cannot be `True` at the same time: the Idle and AwaitingDiscard steady states are different states. This makes the current state clear when the controller reads the CR. `DualPassword` has a different role: it shows whether MySQL currently has a secondary password. It is named after the state it describes, like the `MemoryPressure` condition, rather than describing whether the system is healthy.

The controller calculates the current step from these three conditions and the generation values. It therefore does not need to store a separate `Phase` field in the CR or make an additional API request at runtime. This follows the Kubernetes convention of using Conditions to describe resource state.
```

arosh:
@shunki-fujita 
Add the following rule, please.

```shell
mkdir -p .claude/rules
cat > .claude/rules/markdown.instructions.md <<'EOF'
---
paths:
  - "**/*.md"
---

When writing in English, use CEFR B2-level language: be clear and direct, use common vocabulary, and avoid unnecessarily complex sentence structures.
EOF
```

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,629 @@
...
+- `discardGeneration` must equal `0` (discard must be requested via update once the CR reaches the awaiting-discard steady state — a non-zero value at create time would skip `AwaitingRollout` and the verification window).
+
+**ValidateUpdate:**
+- Both `rotationGeneration` and `discardGeneration` must be monotonically non-decreasing.
+- `discardGeneration <= rotationGeneration`.
+- `rotationGeneration` may only increase while `oldCR.IsIdle()` (Step is `Idle` or `RotationRefused` — i.e. nothing was mutated, so a retry is safe). A previously-stuck cycle (`Blocked` / `Stale`) must be cleared via the recovery procedure (delete + recreate) before a new request.
+- `rotationGeneration` increase additionally requires `cluster.Spec.Replicas > 0` (rechecked at apply time; the controller still re-checks at reconcile time to handle scale-downs after admission).
+- `discardGeneration` may only increase while `oldCR.IsAwaitingDiscard()` (Step is `AwaitingDiscard`; the post-distribute rollout has settled).
+
+**ValidateDelete** — allow:
```
arosh: ` — allow`?

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,629 @@
...
+
+The current step is derived from these three conditions plus the generation comparisons, so the lack of a stored Phase keeps the CR convention-compliant without requiring extra API I/O at runtime.
+
+### Validation Webhook
+
+**ValidateCreate:**
+- The target MySQLCluster (same name, same namespace) must exist.
+- `cluster.Spec.Replicas` must be `> 0`.
+- `rotationGeneration` must equal `1` (counter starts at 1 by convention).
+- `discardGeneration` must equal `0` (discard must be requested via update once the CR reaches the awaiting-discard steady state — a non-zero value at create time would skip `AwaitingRollout` and the verification window).
```
arosh:
```suggestion
**ValidateCreate:**

All of the following conditions are required (AND).

- The target MySQLCluster (same name, same namespace) must exist.
- `cluster.Spec.Replicas` must be `> 0`.
- `rotationGeneration` must equal `1` (counter starts at 1 by convention).
- `discardGeneration` must equal `0`. Discard must be requested via update after the CR reaches the awaiting-discard steady state. A non-zero value at create time would skip `AwaitingRollout` and the verification window.
```

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,629 @@
...
+
+**ValidateUpdate:**
+- Both `rotationGeneration` and `discardGeneration` must be monotonically non-decreasing.
+- `discardGeneration <= rotationGeneration`.
+- `rotationGeneration` may only increase while `oldCR.IsIdle()` (Step is `Idle` or `RotationRefused` — i.e. nothing was mutated, so a retry is safe). A previously-stuck cycle (`Blocked` / `Stale`) must be cleared via the recovery procedure (delete + recreate) before a new request.
+- `rotationGeneration` increase additionally requires `cluster.Spec.Replicas > 0` (rechecked at apply time; the controller still re-checks at reconcile time to handle scale-downs after admission).
+- `discardGeneration` may only increase while `oldCR.IsAwaitingDiscard()` (Step is `AwaitingDiscard`; the post-distribute rollout has settled).
```
arosh:
```suggestion
**ValidateUpdate:**

The following invariant checks apply to every update, regardless of which fields changed (AND):

- `rotationGeneration` must be greater than or equal to its old value.
- `discardGeneration` must be greater than or equal to its old value.
- `discardGeneration <= rotationGeneration`.

In addition, the following condition applies when the corresponding generation is increased:

- If `rotationGeneration` increases, both of these conditions are required (AND):
  - `oldCR.IsIdle()` is true. This means the old Step is `Idle` or `RotationRefused`; nothing was mutated, so a retry is safe. A previously stuck cycle (`Blocked` / `Stale`) must be cleared through the recovery procedure (delete + recreate) before a new request.
  - The live target MySQLCluster has `cluster.Spec.Replicas > 0`. This is rechecked when the request is applied; the controller also re-checks it at reconcile time to handle scale-downs after admission.
- If `discardGeneration` increases, `oldCR.IsAwaitingDiscard()` must be true. The old Step must be `AwaitingDiscard`, which means the post-distribute rollout has settled and the verification window is open.
```

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,629 @@
...
+
+**ValidateDelete** — allow:
+- `cr.IsDeletable()` is true (Step is one of `Idle`, `RotationRefused`, `RotationBlocked`, `DiscardBlocked`, `StalePending`). The `Blocked` / `Stale` cases are the documented recovery escape hatch.
+- The owning MySQLCluster is `NotFound` (GC after owner deletion).
+- The owning MySQLCluster has `DeletionTimestamp` set (`blockOwnerDeletion=true` would otherwise stall cluster termination).
+- The CR carries a stale MySQLCluster ownerRef whose UID does not match the live cluster (recreated cluster; see [Stale CR Handling](#stale-cr-handling-cluster-recreated-under-the-same-name)).
```
arosh:
```suggestion
**ValidateDelete:**

Any one of the following conditions allows deletion (OR).

- `cr.IsDeletable()` is true. This means the Step is one of `Idle`, `RotationRefused`, `RotationBlocked`, `DiscardBlocked`, or `StalePending`. `RotationBlocked`, `DiscardBlocked`, and `StalePending` are the documented recovery escape hatches.
- The owning MySQLCluster is `NotFound` (garbage collection after owner deletion).
- The owning MySQLCluster has `DeletionTimestamp` set. `blockOwnerDeletion=true` would otherwise stall cluster termination.
- The CR carries a stale MySQLCluster ownerRef whose UID does not match the live cluster (the cluster was recreated under the same name; see [Stale CR Handling](#stale-cr-handling-cluster-recreated-under-the-same-name)).
```

docs/kubectl-moco.md
```diff
@@ -57,15 +57,35 @@ To run `mysql` interactively for the instance 2 in `mycluster` in the default na
 $ kubectl moco mysql --index 2 -it mycluster
 ```
 
-## `kubectl moco credential [options] CLUSTER_NAME`
+## `kubectl moco credential`
 
-Fetch the credential information of a specified user
+Manage MySQL credentials for a MOCO cluster.
+
+When called with a cluster name and no subcommand (e.g. `kubectl moco credential CLUSTER_NAME`), it shows the credential for backward compatibility.
```
arosh: @shunki-fujita 
サブコマンドにすることにこだわりをお持ちなのは理解するものの、 `CLUSTER_NAME=rotate` や `CLUSTER_NAME=discard` があると後方互換性を破壊するので、この変更については冷静に考え直したいです。また、サブコマンドの実装がどんどん重厚でメンテ困難なものになってきていると感じています。

以下のようにすることはできますか？

* `kubectl moco credential`: クレデンシャルを得る（これまで通り）
* `kubectl moco rotate-credential CLUSTER_NAME`: Rotate system user passwords for a MOCO cluster
* `kubectl moco discard-old-credential CLUSTER_NAME`: Discard old passwords after a successful credential rotation.

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,644 @@
...
+
+| # | Action | Persistence |
+|---|---|---|
+| 1 | Pre-check: every instance is scanned for pre-existing dual passwords. Skipped if the `RETAIN_STARTED` marker is set (crash recovery). | — |
+| 2 | Set `RETAIN_STARTED` marker (rotationID) in the source Secret. | Secret.Update |
+| 3 | For each instance: connect with the current password, disable `super_read_only` on every instance that runs with it on (replicas, and the intermediate primary when `spec.replicationSourceSecretName` is set), execute `ALTER USER ... RETAIN CURRENT PASSWORD` per user (skipping users where `HasDualPassword` is already true), restore `super_read_only`. | MySQL |
+| 4 | Set `DualPassword=True/Retained` (and clear any prior `RotationReady=False/Blocked` back to `False/Pending`). | Status.Update |
+
+**Pre-check + `RETAIN_STARTED` marker.** If any instance already has a dual-password set from outside this cycle, a `DualPasswordExists` Warning Event is emitted and the step waits. Once the pre-check passes, the marker is persisted so that, after a crash and restart, the reconcile skips the pre-check and resumes RETAIN. From that point, per-user `HasDualPassword` provides idempotency.
```
arosh:
```suggestion
**Pre-check + `RETAIN_STARTED` marker.** If any instance already has a dual-password set from outside this cycle, a `DualPasswordExists` Warning Event is emitted and the step waits. Once the pre-check passes, the marker is persisted so that, after a crash and restart, the reconciliation skips the pre-check and resumes RETAIN. From that point, per-user `HasDualPassword` provides idempotency.
```

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,644 @@
...
+| `kubectl moco credential <cluster>` | Read the per-namespace user Secret (unchanged from previous releases). |
+
+`kubectl get credentialrotation` prints `ROTATIONREADY` / `DISCARDREADY` / `DUALPASSWORD` (the three condition statuses) plus the four generation columns and `AGE`.
+
+### GitOps / ArgoCD
+
+The CR is long-lived and purely declarative, so it works naturally with GitOps. The lifecycle is driven by committing `rotationGeneration` / `discardGeneration` bumps; each commit triggers an ArgoCD sync that advances the cycle. No imperative CLI calls or CR deletions are required for normal operation.
+
+**Do not mix GitOps with `kubectl moco rotate-credential` / `discard-old-credential`.** The CLI patches the same spec fields GitOps manages. If the CLI bumps a counter, GitOps reconcile will try to roll it back, but the webhook rejects any decrease — leaving the resource permanently `OutOfSync`. Worse, the CLI-triggered phase already mutates MySQL passwords irreversibly. Pick one source of truth per environment.
```
arosh:
```suggestion
**Do not mix GitOps with `kubectl moco rotate-credential` / `discard-old-credential`.** The CLI patches the same spec fields GitOps manages. If the CLI bumps a counter, GitOps reconciliation will try to roll it back, but the webhook rejects any decrease — leaving the resource permanently `OutOfSync`. Worse, the CLI-triggered phase already mutates MySQL passwords irreversibly. Pick one source of truth per environment.
```

docs/designdoc/credential_rotation_crd.md
arosh, line 1: @shunki-fujita 
"source Secret" seems to be a new term in this PR, but MOCO already uses "replication source Secret" for the donor connection Secret specified by replicationSourceSecretName. These are different objects: this Secret is the MOCO-managed Secret in the controller namespace that stores the system-user credentials and acts as the input for the per-cluster user and my.cnf Secrets.

Could we call it "controller Secret" consistently, matching `ControllerSecretName()` and the existing comments? In user-facing documentation, "controller-managed credential Secret" could be introduced once and then shortened to "controller Secret".

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,646 @@
...
+
+### Reconciler: Idle → ApplyingRetain
+
+Triggered when the outstanding phase is `rotation` (i.e. `spec.rotationGeneration > status.observedRotationGeneration`) and the CR is in `Step ∈ {Idle, RotationRefused, RotationBlocked}` with `cluster.Spec.Replicas > 0`.
+
+| # | Action | Persistence |
+|---|---|---|
+| 1 | Generate (or reuse on crash recovery) the rotation UUID. | — |
+| 2 | Write 8 `*_PENDING` keys and `ROTATION_ID` into the controller Secret. | Secret.Update |
+| 3 | Set `RotationReady=False/Pending`, `DiscardReady=False/Pending`, `DualPassword=False/NotRetained`. | Status.Update |
```
arosh:
```suggestion
| 3 | Set `RotationReady=False` (`reason=Pending`), `DiscardReady=False` (`reason=Pending`), `DualPassword=False` (`reason=NotRetained`). | Status.Update |
```

他の箇所も同様の表記にしてください

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,646 @@
...
+
+`kubectl get credentialrotation` prints `ROTATIONREADY` / `DISCARDREADY` / `DUALPASSWORD` (the three condition statuses) plus the four generation columns and `AGE`.
+
+### GitOps / ArgoCD
+
+The CR is long-lived and purely declarative, so it works naturally with GitOps. The lifecycle is driven by committing `rotationGeneration` / `discardGeneration` bumps; each commit triggers an ArgoCD sync that advances the cycle. No imperative CLI calls or CR deletions are required for normal operation.
+
+**Do not mix GitOps with `kubectl moco rotate-credential` / `discard-old-credential`.** The CLI patches the same spec fields GitOps manages. If the CLI bumps a counter, GitOps reconciliation will try to roll it back, but the webhook rejects any decrease — leaving the resource permanently `OutOfSync`. Worse, the CLI-triggered phase already mutates MySQL passwords irreversibly. Pick one source of truth per environment.
+
+## Rotate
```
arosh:
```suggestion
## Rotation Phase
```

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,646 @@
...
+
+### Scaled-down clusters (replicas=0)
+
+A cluster with 0 replicas stops rotation at three points:
+- At admission: the webhook rejects CR creation or a `rotationGeneration` bump when `cluster.Spec.Replicas <= 0`.
+- At reconcile time, before any mutation: if the cluster was scaled down to 0 after admission, `handleStartRotation` emits a `RotationRefused` Warning Event and sets `RotationReady=False/Refused`. Nothing has been mutated; the CR stays in `Step=RotationRefused` and remains eligible for retry.
+- If the cluster is scaled down to 0 *after* pending passwords were written, the ClusterManager handler emits a `RotationBlocked` Warning Event and sets `RotationReady=False/Blocked`. Recovery requires either scaling the cluster back up (the reconciler resumes automatically when it sees a healthy cluster again) or following the recovery procedure.
+
+## Discard
```
arosh:
```suggestion
## Discard Phase
```

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,646 @@
...
+Triggered when the outstanding phase is `rotation` (i.e. `spec.rotationGeneration > status.observedRotationGeneration`) and the CR is in `Step ∈ {Idle, RotationRefused, RotationBlocked}` with `cluster.Spec.Replicas > 0`.
+
+| # | Action | Persistence |
+|---|---|---|
+| 1 | Generate (or reuse on crash recovery) the rotation UUID. | — |
+| 2 | Write 8 `*_PENDING` keys and `ROTATION_ID` into the controller Secret. | Secret.Update |
+| 3 | Set `RotationReady=False/Pending`, `DiscardReady=False/Pending`, `DualPassword=False/NotRetained`. | Status.Update |
+
+### ClusterManager: ApplyingRetain → DistributingPassword
+
```
arosh, line 445: Describe "Triggered when ..."

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,646 @@
...
+| # | Action | Persistence |
+|---|---|---|
+| 1 | Pre-check: every instance is scanned for pre-existing dual passwords. Skipped if the `RETAIN_STARTED` marker is set (crash recovery). | — |
+| 2 | Set `RETAIN_STARTED` marker (rotationID) in the controller Secret. | Secret.Update |
+| 3 | For each instance: connect with the current password, disable `super_read_only` on every instance that runs with it on (replicas, and the intermediate primary when `spec.replicationSourceSecretName` is set), execute `ALTER USER ... RETAIN CURRENT PASSWORD` per user (skipping users where `HasDualPassword` is already true), restore `super_read_only`. | MySQL |
```
arosh:
```suggestion
| 3 | For each instance, connect with the current password. If the instance is a replica, or the intermediate primary when `spec.replicationSourceSecretName` is set, disable `super_read_only` if it is enabled. Run `ALTER USER ... RETAIN CURRENT PASSWORD` for each user, skipping users that already have a dual password. Restore `super_read_only` after the updates. | MySQL |
```

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,646 @@
...
+| 4 | Set `DualPassword=True/Retained` (and clear any prior `RotationReady=False/Blocked` back to `False/Pending`). | Status.Update |
+
+**Pre-check + `RETAIN_STARTED` marker.** If any instance already has a dual-password set from outside this cycle, a `DualPasswordExists` Warning Event is emitted and the step waits. Once the pre-check passes, the marker is persisted so that, after a crash and restart, the reconciliation skips the pre-check and resumes RETAIN. From that point, per-user `HasDualPassword` provides idempotency.
```
arosh:
```suggestion
#### Pre-check and crash recovery

If any instance already has a dual password from outside this rotation cycle, emit a `DualPasswordExists` Warning Event and wait. After the pre-check succeeds, store the marker before running RETAIN. If the controller crashes and restarts, the marker tells the reconciliation to skip the pre-check and resume RETAIN. The `HasDualPassword` check for each user makes retries safe and idempotent.
```

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,646 @@
...
+| 3 | Set `RotationReady=False/Pending`, `DiscardReady=False/Pending`, `DualPassword=False/NotRetained`. | Status.Update |
+
+### ClusterManager: ApplyingRetain → DistributingPassword
+
+| # | Action | Persistence |
+|---|---|---|
+| 1 | Pre-check: every instance is scanned for pre-existing dual passwords. Skipped if the `RETAIN_STARTED` marker is set (crash recovery). | — |
+| 2 | Set `RETAIN_STARTED` marker (rotationID) in the controller Secret. | Secret.Update |
+| 3 | For each instance: connect with the current password, disable `super_read_only` on every instance that runs with it on (replicas, and the intermediate primary when `spec.replicationSourceSecretName` is set), execute `ALTER USER ... RETAIN CURRENT PASSWORD` per user (skipping users where `HasDualPassword` is already true), restore `super_read_only`. | MySQL |
+| 4 | Set `DualPassword=True/Retained` (and clear any prior `RotationReady=False/Blocked` back to `False/Pending`). | Status.Update |
```
arosh:
```suggestion
| 4 | Set `DualPassword=True` (`reason=Retained`). If `RotationReady` was previously `False` (`reason=Blocked`), reset it to `False` (`reason=Pending`). | Status.Update |
```

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,646 @@
...
+
+**Pre-check + `RETAIN_STARTED` marker.** If any instance already has a dual-password set from outside this cycle, a `DualPasswordExists` Warning Event is emitted and the step waits. Once the pre-check passes, the marker is persisted so that, after a crash and restart, the reconciliation skips the pre-check and resumes RETAIN. From that point, per-user `HasDualPassword` provides idempotency.
+
+### Reconciler: DistributingPassword → AwaitingRollout
+
+| # | Action | Persistence |
+|---|---|---|
+| 1 | Apply pending passwords to per-namespace user Secret and my.cnf Secret. | Secret.Apply |
+| 2 | Add `moco.cybozu.com/password-rotation-restart: <rotationID>` to the StatefulSet pod template via SSA under field manager `moco-credential-rotation` with `ForceOwnership`. | StatefulSet.Apply |
+| 3 | Promote `observedRotationGeneration = spec.rotationGeneration`. `DiscardReady` stays `False/Pending`. | Status.Update |
```
arosh:
```suggestion
| 3 | Record the rotation as complete by setting `status.observedRotationGeneration` to `spec.rotationGeneration`. Keep `DiscardReady=False/Pending` until the StatefulSet rollout is complete. | Status.Update |
```

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,646 @@
...
+| 4 | StatefulSet controller rolls the Pods. | — |
+
+### Reconciler: AwaitingRollout → AwaitingDiscard
+
+| # | Action | Persistence |
+|---|---|---|
+| 1 | Get the StatefulSet; if `NotFound`, requeue. | — |
+| 2 | Check rollout completion (`ObservedGeneration` caught up, `CurrentRevision == UpdateRevision`, `UpdatedReplicas == Replicas`, `ReadyReplicas == Replicas`). If in flight, requeue. | — |
```
arosh:
```suggestion
| 2 | Check whether the StatefulSet rollout is complete. Confirm that `status.observedGeneration` has caught up with `metadata.generation`, `status.currentRevision` matches `status.updateRevision`, and `status.replicas`, `status.updatedReplicas`, and `status.readyReplicas` all equal the desired `spec.replicas`. If any check is not satisfied, do not mark the rollout complete; requeue and check again later. | — |
```

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,646 @@
...
+| 3 | Once settled: `DiscardReady=True/Reconciled` and emit `AwaitingDiscard` Event. | Status.Update |
```
arosh:
```suggestion
| 3 | After all rollout checks pass, set the `DiscardReady` condition in `CredentialRotation.status` to `True` with reason `Reconciled`. This records that the new password has been distributed and all Pods are ready, so the verification window is open. Emit an `AwaitingDiscard` Event to report that the operator may request the discard step. | Status.Update |
```

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,646 @@
...
+
+**Why wait for the rollout here, not inside DISCARD?** The verification window only makes sense once every Pod is using the new password. Setting `DiscardReady=True` earlier would let `kubectl wait --for=condition=DiscardReady` return while the rollout is still in flight, and would let an automation script start `discard` while some Pods still depend on the old password — DISCARD at that point would remove the secondary password those Pods still use. The rollout is also a K8s concern, so it belongs in the Reconciler.
```
arosh:
```suggestion
#### Why the reconciler waits for the rollout

The verification window is safe only after every Pod has restarted and its MySQL clients have picked up the pending password from the distributed Secrets. For example, `moco-agent` reads credentials from the per-namespace user Secret, and `mysqld-exporter` reads them from the `my.cnf` Secret. If the controller set `DiscardReady=True` before the rollout finished, `kubectl wait --for=condition=DiscardReady` could return while some Pods' clients still used the old password. An automation script could then start the discard step. `DISCARD OLD PASSWORD` would remove the secondary password that those clients still need, and they could lose access to MySQL.

The rollout is a Kubernetes operation, so the CredentialRotation Reconciler checks the StatefulSet status and opens the verification window. The ClusterManager performs the MySQL `DISCARD` operation only after this check has succeeded.
```

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,646 @@
...
+
+### Scaled-down clusters (replicas=0)
+
+A cluster with 0 replicas stops rotation at three points:
+- At admission: the webhook rejects CR creation or a `rotationGeneration` bump when `cluster.Spec.Replicas <= 0`.
+- At reconcile time, before any mutation: if the cluster was scaled down to 0 after admission, `handleStartRotation` emits a `RotationRefused` Warning Event and sets `RotationReady=False/Refused`. Nothing has been mutated; the CR stays in `Step=RotationRefused` and remains eligible for retry.
+- If the cluster is scaled down to 0 *after* pending passwords were written, the ClusterManager handler emits a `RotationBlocked` Warning Event and sets `RotationReady=False/Blocked`. Recovery requires either scaling the cluster back up (the reconciler resumes automatically when it sees a healthy cluster again) or following the recovery procedure.
```
arosh: このセクションってこの位置に置かれている必要がありますか？

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,646 @@
...
+| `ClusterManager.handlePasswordRotation` | Return early; do not run RETAIN / DISCARD |
+| `MySQLClusterReconciler.reconcileV1Secret` | Ignore the CR; distribute current passwords normally |
+| `kubectl moco rotate-credential` / `discard-old-credential` | Refuse with an error instructing the user to delete the stale CR |
+
+"Stale" means the CR has a MySQLCluster ownerReference whose UID does **not** match the live cluster, with no matching reference. A CR with no MySQLCluster ownerReference yet (just-created, not yet adopted) is treated as **fresh**.
+
+## Assumptions
+
+- No MOCO system user has a dual password when rotation starts. The pre-check on `ApplyingRetain` validates this; on violation it emits a `DualPasswordExists` Warning Event and waits. See [Recovery: Dual Password Exists While No Active Rotation](#dual-password-exists-while-no-active-rotation).
+- MySQL version is 8.0.14+ (dual password support).
```
arosh: このセクションは冒頭に書かれていないといけないと思います

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,646 @@
...
+
+## Discard
+
+### Reconciler: AwaitingDiscard → ApplyingDiscard (initial transition)
+
+| # | Action | Persistence |
+|---|---|---|
+| 1 | Validate `spec.discardGeneration > status.observedDiscardGeneration`. | — |
+| 2 | If `cluster.Spec.Replicas <= 0`: emit `DiscardRefused` Warning Event and set `DiscardReady=False/Refused`. The webhook forbids reverting `discardGeneration` once bumped, so a Refused state at this point simply waits for the cluster to be scaled up. | Status.Update |
+| 3 | Whenever `DiscardReady` is not already `False/Pending` (initial bump, or recovery from `False/Refused` or `False/Blocked`): set `DiscardReady=False/Pending`, emit `DiscardStarted` Event, requeue. Once it is `False/Pending`, subsequent reconciles just requeue while ClusterManager drives DISCARD. | Status.Update |
```
arosh: - "1. Validate spec.discardGeneration > status.observedDiscardGeneration." は step の中の procedure ではありません。step に入るかどうかを決める trigger です
- 

```suggestion
### Reconciler: AwaitingDiscard → ApplyingDiscard (initial transition)

Triggered when the outstanding phase is `discard` (i.e. `spec.discardGeneration > status.observedDiscardGeneration`) and the CR is in `Step ∈ {AwaitingDiscard, DiscardRefused, DiscardBlocked}`.

| # | Action | Persistence |
|---|---|---|
| 1 | If `cluster.Spec.Replicas <= 0`: emit `DiscardRefused` Warning Event and set `DiscardReady=False/Refused`. The webhook forbids reverting `discardGeneration` once bumped, so a Refused state at this point simply waits for the cluster to be scaled up. | Status.Update |
| 2 | Whenever `DiscardReady` is not already `False/Pending` (initial bump, or recovery from `False/Refused` or `False/Blocked`): set `DiscardReady=False/Pending`, emit `DiscardStarted` Event, requeue. Once it is `False/Pending`, subsequent reconciles just requeue while ClusterManager drives DISCARD. | Status.Update |
```

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,646 @@
...
+### ClusterManager: ApplyingDiscard → Finalizing
+
+| # | Action | Persistence |
+|---|---|---|
+| 1 | Handshake: wait until `DiscardReady=False/Pending`. While the condition is `True`, `False/Refused`, or `False/Blocked`, return early. | — |
+| 2 | Determine the target auth plugin via `GetAuthPlugin` on the primary. | MySQL (read-only) |
+| 3 | For each instance: connect with the **pending** password, disable `super_read_only` on every instance that runs with it on (replicas, and the intermediate primary when `spec.replicationSourceSecretName` is set), execute `DISCARD OLD PASSWORD` per user (skipped where `HasDualPassword` is already false), then auth plugin migration. Restore `super_read_only`. | MySQL |
+| 4 | Set `DualPassword=False/NotRetained`. | Status.Update |
+
+**Why the handshake?** Both Reconciler and ClusterManager observe `Step=ApplyingDiscard` once the operator bumps `discardGeneration`. Without the handshake, ClusterManager could race ahead and run DISCARD before the Reconciler flips `DiscardReady` from `True/Reconciled` to `False/Pending`, skipping the `DiscardStarted` Event.
```
arosh: * handshake とかいう新しい語彙を導入する必要は無いと思います。CredentialRotationReconciler による discard request を意味する DiscardReady=False/Pending が観測されるまで起動しないと言えば十分です

```suggestion
### ClusterManager: ApplyingDiscard → Finalizing

Triggered on a ClusterManager tick when `cr.Step() == StepApplyingDiscard` (that is, `spec.discardGeneration > status.observedDiscardGeneration` and `DualPassword=True`). Before executing SQL, ClusterManager requires `DiscardReady=False/Pending`, which is written by the CredentialRotationReconciler after the discard request is observed. If the condition is not yet present, ClusterManager skips rotation work for that tick and retries on a later tick. This ordering records the in-flight state and emits the `DiscardStarted` Event before DISCARD SQL; it is not a blocking wait or a rollout gate.

| # | Action | Persistence |
|---|---|---|
| 1 | Determine the target auth plugin via `GetAuthPlugin` on the primary. | MySQL (read-only) |
| 2 | For each instance: connect with the **pending** password, disable `super_read_only` on every instance that runs with it on (replicas, and the intermediate primary when `spec.replicationSourceSecretName` is set), execute `DISCARD OLD PASSWORD` per user (skipped where `HasDualPassword` is already false), then auth plugin migration. Restore `super_read_only`. | MySQL |
| 3 | Set `DualPassword=False/NotRetained`. | Status.Update |
```

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,646 @@
...
+
+**Why connect with the pending password?** DISCARD removes the old password. Connecting with the old password would fail immediately after DISCARD succeeds. Using the pending password also confirms that distribution succeeded.
+
+**No rollout re-wait.** The post-distribute rollout is already gated by `DiscardReady=True` (set in `AwaitingRollout`), so by the time the operator can bump `discardGeneration` every Pod is already running with the new password.
```
arosh: ここは現在議論中の「pending password を current password に昇格する時期」の件の行方によっては削除できる

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,646 @@
...
+### Scaled-down clusters (replicas=0)
+
+A cluster with 0 replicas stops rotation at three points:
+- At admission: the webhook rejects CR creation or a `rotationGeneration` bump when `cluster.Spec.Replicas <= 0`.
+- At reconcile time, before any mutation: if the cluster was scaled down to 0 after admission, `handleStartRotation` emits a `RotationRefused` Warning Event and sets `RotationReady=False/Refused`. Nothing has been mutated; the CR stays in `Step=RotationRefused` and remains eligible for retry.
+- If the cluster is scaled down to 0 *after* pending passwords were written, the ClusterManager handler emits a `RotationBlocked` Warning Event and sets `RotationReady=False/Blocked`. Recovery requires either scaling the cluster back up (the reconciler resumes automatically when it sees a healthy cluster again) or following the recovery procedure.
+
+## Discard
+
+### Reconciler: AwaitingDiscard → ApplyingDiscard (initial transition)
```
arosh: Could you explain why `(initial transition)` should be described all the way?

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,646 @@
...
+1. **Crash safety of the promote step.** `PromotePendingPasswords` promotes pending passwords to current by renaming keys within a single object — atomic at the Secret level. A separate Secret would require copying data between two objects; a crash between the read and the write could lose the new passwords permanently.
+2. **Simpler failure modes.** With a single Secret, the only question on crash recovery is "did the update succeed?". With two Secrets, every sub-step has to reason about cross-object consistency.
+3. **Idempotency.** `SetPendingPasswords` checks if matching pending keys already exist; `PromotePendingPasswords` is a no-op when no pending keys remain. Both work on a single object.
+
+## Component Details
+
+### CredentialRotationReconciler (new)
+
+The reconciler watches:
+- `CredentialRotation` (primary).
```
arosh:
```suggestion
- `CredentialRotation` (primary watched resource).
```

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,646 @@
...
+- `MySQLCluster` (filtered to `Spec.Replicas` change or `DeletionTimestamp` flip), mapped to the same namespace/name — so a `Refused` / `Blocked` cycle resumes immediately on scale-up.
+- `Secret` in the system namespace (filtered by the `mysql-<ns>.<name>` naming pattern), so a Stale controller Secret can be cleaned up without waiting for the 15-second requeue.
+
+For sub-steps owned by ClusterManager (`ApplyingRetain`, `ApplyingDiscard` DB work), the reconciler requeues every 15s while observing the condition for progress.
+
+Server-Side Apply writes for the rolling-restart annotation use the dedicated field manager `moco-credential-rotation` with `ForceOwnership`, so the annotation is not removed by `MySQLClusterReconciler`'s `moco-controller` field manager on the next cluster reconcile.
```
arosh:
```suggestion
The rolling-restart annotation is applied with the dedicated field manager `moco-credential-rotation` and `ForceOwnership`. This keeps the annotation owned by the credential rotation reconciler, so the regular `MySQLClusterReconciler` does not remove it during its next reconcile.
```

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,646 @@
...
+
+### ClusterManager
+
+ClusterManager reads the CredentialRotation CR inside each tick and dispatches on `cr.Step()`:
+- `ApplyingRetain` → run the RETAIN flow on this cluster.
+- `ApplyingDiscard` → run the DISCARD flow (blocked on the `DiscardReady=False/Pending` handshake first).
+- Any other step → no-op for rotation; normal clustering continues.
+
+A CR whose ownerReference UID does not match the live cluster (stale CR) is ignored. Status writes use `retry.RetryOnConflict` with a fresh `Get` inside the retry to avoid conflicts with concurrent status updates from the reconciler.
```
arosh: これは Kubernetes controller を書こうとしている人にとっては当たり前の話で、デザインドキュメントに書くレベルのことではないと思います。デザインドキュメントに「クエリを書くときは SQL インジェクションに気を付ける」とかわざわざ書かないでしょう

```suggestion
A CR whose ownerReference UID does not match the live cluster (stale CR) is ignored.
```

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,646 @@
...
+
+`AwaitingRollout`, `AwaitingDiscard`, and `DiscardRefused` are **not** deletable: MySQL still holds dual passwords. Operators must scale the cluster down first (which transitions the CR to `RotationBlocked` / `DiscardBlocked`).
+
+The CR does **not** use a finalizer for automatic rollback: rollback requires connecting to every MySQL instance (which may not be possible during deletion), and a partial rollback is worse than no rollback.
+
+### MySQLCluster deletion
+
+The ownerReference (`blockOwnerDeletion=true`) means GC must delete the CR before the MySQLCluster finishes terminating. The webhook explicitly allows delete when the owning cluster has a non-nil `DeletionTimestamp`, even for in-flight steps; otherwise the cluster would be stuck in `Terminating` indefinitely. No special teardown is needed — the MySQL instances are being destroyed too.
+
+### Stale CR handling (cluster recreated under the same name)
```
arosh, line 764: 1. 本当に blockOwnerDeletion=true が設定されていますか？ 

https://github.com/cybozu-go/moco/blob/3414da780b2e0dfee4a771c0be53d6d03d47e5c1/controllers/credentialrotation_controller.go#L88-L96

2. `blockOwnerDeletion=true` が設定済みであれば、旧 MySQLCluster は CredentialRotation の削除前に完全削除されず、同名の新しい MySQLCluster も作成できないため、recreate (stale CR?) のハンドリングを頑張るモチベーションが分かりません。コーナーケースとして CredentialRotationReconciler.Reconcile による ownerReferences の設定前に MySQLCluster が recreate された場合は `blockOwnerDeletion=true` が設定されていないですが、これは何か処理が始まる前に設定されて early return されるので問題ないと思います。

clustering/password_rotation.go
```diff
@@ -0,0 +1,466 @@
...
+	if needsSuperReadOnlyOff {
+		if err := op.SetSuperReadOnly(ctx, false); err != nil {
+			return fmt.Errorf("failed to disable super_read_only on instance %d: %w", instanceIndex, err)
+		}
+		defer func() {
+			if err := op.SetSuperReadOnly(ctx, true); err != nil {
+				log.Error(err, "failed to re-enable super_read_only (clustering loop will recover)",
+					"instance", instanceIndex)
+			}
+		}()
```
arosh, line 612: Failure of `op.SetSuperReadOnly` should be error of `rotateInstanceUsers`. Use named return values.

clustering/password_rotation.go
```diff
@@ -0,0 +1,466 @@
...
+	if needsSuperReadOnlyOff {
+		if err := op.SetSuperReadOnly(ctx, false); err != nil {
+			return fmt.Errorf("failed to disable super_read_only on instance %d for discard: %w", instanceIndex, err)
+		}
+		defer func() {
+			if err := op.SetSuperReadOnly(ctx, true); err != nil {
+				log.Error(err, "failed to re-enable super_read_only (clustering loop will recover)",
+					"instance", instanceIndex)
+			}
+		}()
```
arosh, line 673: Failure of `op.SetSuperReadOnly` should be error of `discardInstanceUsers`. Use named return values.

controllers/credentialrotation_controller.go
```diff
@@ -0,0 +1,686 @@
...
+	return ctrl.Result{RequeueAfter: credRotationRequeueInterval}, nil
+}
+
+// handleAwaitingRollout waits for the post-distribute StatefulSet rollout
+// to settle. While the rollout is in flight, some Pods may still be
+// running with the old password loaded via EnvFrom; flipping
+// DiscardReady=True before every Pod has picked up the new password
+// would let an operator-initiated DISCARD strip the secondary password
+// out from under them. Once the rollout is complete, all Pods are using
+// the new password and the verification window is genuinely open.
```
arosh: cached reader の存在を考慮できていますか？ AnnPasswordRotationRestart の書き込み直後に handleAwaitingRollout を呼ぶと、おそらく再起動完了と判定されてしまうと思います。podTemplate の AnnPasswordRotationRestart が期待する RotationID と一致することを条件に入れるべきではないでしょうか

docs/kubectl-moco.md
```diff
@@ -66,6 +66,20 @@ Fetch the credential information of a specified user
...
+
+This can only be run when no rotation cycle is in flight (the CredentialRotation CR is idle: `RotationReady=True`, or the previous request was refused without any changes).
+
+## `kubectl moco discard-old-credential CLUSTER_NAME`
+
+Discard old passwords after a successful credential rotation.
+Bumps `spec.discardGeneration` to match `spec.rotationGeneration` on the CredentialRotation CR.
+
+This can only be run when the CredentialRotation CR is awaiting discard (`DiscardReady=True`; the post-distribute rollout has settled).
+
```
arosh, line 98: 安全のため、

- MySQLCluster が Healthy である
- `spec.offline` が true でない
- `moco.cybozu.com/reconciliation-stopped=true` が設定されていない

くらいはコマンド実行時にチェックしてもよいのではと思います

cmd/kubectl-moco/cmd/rotate_credential.go
```diff
@@ -0,0 +1,124 @@
...
+		"spec": map[string]any{
+			"rotationGeneration": newGen,
+		},
+	})
+	if err != nil {
+		return fmt.Errorf("failed to marshal patch: %w", err)
+	}
+	if err := kubeClient.Patch(ctx, cr, client.RawPatch(types.MergePatchType, patch)); err != nil {
+		return fmt.Errorf("failed to patch CredentialRotation: %w", err)
+	}
```
arosh: 複数の kubectl-moco が動いていた際にロストアップデートが生じます。 .Update のほうがよいのでは？

docs/designdoc/credential_rotation_crd.md
arosh: 未修整です

api/v1beta2/credentialrotation_types.go
```diff
@@ -0,0 +1,209 @@
...
+	// +kubebuilder:validation:MaxLength=36
+	// +optional
+	RotationID string `json:"rotationID,omitempty"`
+
+	// ObservedRotationGeneration is the last rotationGeneration whose
+	// rotation phase (RETAIN + pending-password distribution) completed
+	// successfully. Equality with spec.rotationGeneration is a necessary
+	// condition for the cycle to leave the rotation phase, but not
+	// sufficient on its own: RotationReady=True is only set at the very
+	// end of the full cycle (after the discard phase finalises).
```
arosh: The design document now correctly defines `observedRotationGeneration` as being updated immediately after promotion, before Secret distribution and the StatefulSet rollout. However, the public API comment still says that it represents completion of “RETAIN + pending-password distribution”, which describes the old workflow.

Please update the comment directly above `CredentialRotationStatus.ObservedRotationGeneration` in `api/v1beta2/credentialrotation_types.go` to state that it is updated after RETAIN succeeds on every instance and the pending password is promoted to the controller Secret's current password. Please also regenerate `docs/crd_credentialrotation_v1beta2.md` so the generated API reference has the same meaning.

`observedRotationGeneration == spec.rotationGeneration` should not imply that Secret distribution, the StatefulSet rollout, or the full rotate-discard cycle has completed.

docs/designdoc/credential_rotation_crd.md
```diff
@@ -0,0 +1,759 @@
...
+  │  Finalizing (DualPw=False)                                               │
+  │    │ Reconciler: delete *_OLD keys and ROTATION_ID (cleanup only —       │
+  │    │            no password movement); promote observedDiscardGen;       │
+  │    │            RotationReady→True (back to Idle)                        │
+  │    ▼                                                                     │
+  │  Idle                                                                    │
+  └──────────────────────────────────────────────────────────────────────────┘
+```
+
+> In this diagram, `Cond→False/Pending` is compact notation for setting the condition to `status: False` with `reason: Pending`. The prose and tables in this document spell it out as `Cond=False` (`reason=Pending`).
```
arosh: ずいぶん難しい英単語をご存じなんですね

```suggestion
> In this diagram, `Cond→False/Pending` is compact notation for setting the condition to `status: False` with `reason: Pending`. The text and tables in this document use `Cond=False` (`reason=Pending`) instead.
```

pkg/dbop/test_util.go
```diff
@@ -200,6 +200,8 @@ func newTestOperator(cluster *mocov1beta2.MySQLCluster, pwd *password.MySQLPassw
 		passwd:    pwd,
 		index:     index,
 		db:        udb,
+		// The Docker port mapping publishes the normal MySQL port.
+		verifyAddr: fmt.Sprintf("localhost:%d", port),
```
arosh: Use `net.joinHostPort`

pkg/dbop/password_rotation_test.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/cybozu-go/moco/pkg/constants"
+	"github.com/cybozu-go/moco/pkg/password"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("password rotation", func() {
+	ctx := context.Background()
+
+	It("should rotate, verify, discard, and migrate system user passwords", func() {
```
arosh:
```suggestion
	It("should rotate, verify, discard, and migrate system user passwords", func(ctx SpecContext) {
```

ほかにも類似の箇所があれば直してください

docs/crd_credentialrotation_v1beta2.md
```diff
@@ -0,0 +1,58 @@
...
+
+[Back to Custom Resources](#custom-resources)
+
+#### CredentialRotationStatus
+
+CredentialRotationStatus defines the observed state of CredentialRotation.
+
+| Field | Description | Scheme | Required |
+| ----- | ----------- | ------ | -------- |
+| observedGeneration | ObservedGeneration reflects the .metadata.generation that the controller has most recently reconciled. | int64 | true |
```
arosh: @shunki-fujita 
`credentialrotation_types.go` says ObservedGeneration is **optional**, whereas crd_credentialrotation_v1beta2.md says observedGeneration is **required**.

https://github.com/cybozu-go/moco/blob/48e2d26ac0288a9931498266c75437f603a4aa22/api/v1beta2/credentialrotation_types.go#L76-L81

Could you explain why such mismatch happend? crd_credentialrotation_v1beta2.md is automatically generated, isn't it?

shunki-fujita: The doc is auto-generated, but the two generators look at different sources: controller-gen derives the CRD schema's optionality from the +optional marker, while crd-to-markdown derives the Required column solely from the presence of omitempty in the JSON tag.

shunki-fujita: It appears that a similar issue is occurring here as well:

https://github.com/cybozu-go/moco/blob/8840e4cf410b7ea3e86d87e9421ff6762ac854c0/docs/crd_mysqlcluster_v1beta2.md#mysqlclusterstatus

- status.backup
- status.reconcileInfo
- status.backup.uuidSet
- status.reconcileInfo.reconcileVersion

https://github.com/cybozu-go/moco/blob/40f54d72d651c9dc58613e07121220a62ebeeb9c/api/v1beta2/mysqlcluster_types.go

PR #924, opened 2026-06-19
Prevent MySQLCluster controller startup slowdown in large clusters

### LINKED ISSUE #922: Prevent moco-controller startup cache sync failures from disabling failover and switchover
## Summary
After upgrading controller-runtime to 0.23.x, moco-controller can fail to start when a cluster has a large number of `MySQLCluster` resources and secondary resources. In production, this caused all moco-controller replicas to become unavailable, which disabled MOCO failover and switchover.

We need to introduce multiple safeguards so that moco-controller remains available during startup, upgrade, and leader failover even in large-scale environments.

## Context
controller-runtime 0.23.x changed controller source startup behavior to wait not only for informer caches to sync, but also for event handlers to process initial objects. This exposed expensive secondary resource handlers in MOCO.

The current `MySQLClusterReconciler` watches secondary resources such as `ConfigMap` and `BackupPolicy`. For each secondary resource event, the handler lists all `MySQLCluster` objects in the namespace and scans them to find matching references. With around 1000 `MySQLCluster` resources, this can make controller startup slow enough to hit cache sync timeout errors such as:

```text
failed to wait for mysqlcluster caches to sync kind source
timed out waiting for cache to be synced for Kind *v1.ConfigMap
```

When this happens during rollout or restart, moco-controller replicas may fail readiness/startup and the controller fleet can become unavailable. Since moco-controller is responsible for MOCO operational workflows, this directly impacts failover and switchover.

Relevant upstream context:

- kubernetes-sigs/controller-runtime#3406 changed source startup to wait for handler sync.
- kubernetes-sigs/controller-runtime#3466 discusses regressions caused by slow handlers during initial sync.
- Warm replica support can reduce leader failover downtime, but it does not by itself reduce the total work done by expensive handlers.

## Plan
- [ ] Add a regression test that demonstrates the current startup failure or severe slowdown with around 1000 `MySQLCluster` resources.
- [ ] Add a configurable `CacheSyncTimeout` flag for moco-controller.
- [ ] Optimize secondary resource handlers in `MySQLClusterReconciler`.
- [ ] Make readiness depend on controller warmup completion during moco-controller RollingUpdate.
- [ ] Tighten `PodWatcher` filtering so it only watches or enqueues MOCO-managed Pods.

### Add a large-scale startup regression test
Introduce a test that reproduces the pathological behavior with many `MySQLCluster` resources and secondary resources.

The test should validate that the current handler pattern does not scale and should protect future changes from reintroducing the same issue. We should evaluate whether this is best implemented with envtest or with a fake client/cache setup. Envtest is likely closer to the real controller-runtime cache behavior, while a fake setup may be faster and easier to run in unit tests.

The test should cover at least the `ConfigMap` watch path, because that is one of the observed timeout kinds. If practical, also cover `BackupPolicy`.

### Add a configurable CacheSyncTimeout flag
Expose a moco-controller flag to configure controller-runtime `Controller.CacheSyncTimeout`.

This is a mitigation, not the primary fix. It gives operators a way to recover or tune behavior in large environments while deeper handler optimizations are rolled out.

The default should preserve current controller-runtime behavior unless we intentionally choose a safer MOCO-specific default.

### Optimize secondary resource handlers
Replace the current list-and-scan handlers with indexed lookups.

The `ConfigMap` handler should not list all `MySQLCluster` objects and scan `spec.mysqlConfigMapName` for every ConfigMap event. Instead, register a field index for `MySQLCluster.spec.mysqlConfigMapName` and use `client.MatchingFields`.

The `BackupPolicy` handler should follow the same pattern with `MySQLCluster.spec.backupPolicyName`.

This should reduce initial handler sync cost from roughly `secondary resource count * MySQLCluster count` to indexed lookups for only matching clusters.

### Gate readiness on warmup completion
During RollingUpdate, a new moco-controller Pod should not become ready before its controllers have completed source/cache warmup. Otherwise Kubernetes may continue the rollout while the replacement replica is not yet capable of taking over failover/switchover duties quickly.

Investigate controller-runtime warm replica support and readiness integration. If `EnableWarmup` is used, readiness should only return OK after the relevant controllers have completed warmup. If controller-runtime does not expose enough state directly, introduce MOCO-side tracking around controller startup/warmup completion.

This should be designed carefully because warm replicas increase API server List/Watch load.

### Tighten PodWatcher filtering
`PodWatcher` currently watches Pods broadly and filters later in reconcile. In large clusters, this can create unnecessary initial events and queue pressure.

Restrict the watch path to MOCO-managed Pods as early as possible, using predicates and/or cache selectors where appropriate. The goal is to avoid processing unrelated Pods during startup and normal operation.

## Note
Warm replicas are useful for reducing downtime during leader failover, but they are not a substitute for optimizing slow event handlers. The primary root cause is the synchronous startup cost of secondary resource handlers after controller-runtime 0.23.x.

arosh:
## Summary
This PR fixes the main cause of issue #922.  
It reduces the startup cost of `moco-controller` when many `MySQLCluster` and secondary resources exist in the same namespace.

The current implementation scanned all `MySQLCluster` objects for each `ConfigMap` and `BackupPolicy` event. This became expensive after the controller-runtime upgrade and could make cache sync fail during controller startup. In that case, MOCO failover and switchover could be unavailable because all controller replicas might fail to become ready.

## Context
After upgrading to controller-runtime 0.23.x, controller startup became more sensitive to slow event handlers during initial cache sync.

In MOCO, the `MySQLClusterReconciler` watches secondary resources such as `ConfigMap` and `BackupPolicy`. The old handlers used a list-and-scan pattern, so startup work increased too much in large-scale environments. This was the root cause behind the cache sync timeout described in issue #922.

## Implementation
This PR replaces the list-and-scan logic with field index lookups.

`MySQLCluster` objects are now indexed by:
- `spec.mysqlConfigMapName`
- `spec.backupPolicyName`

During `ConfigMap` and `BackupPolicy` events, the reconciler now lists only matching `MySQLCluster` objects by using `client.MatchingFields`, instead of scanning every cluster in the namespace.

This PR also adds a startup regression test for the `MySQLCluster` controller. The test creates 1000 `MySQLCluster` objects and 1000 `ConfigMap` objects before the manager starts, then checks that the controller can still start and process initial events in time.

## Note

In addition, test startup goroutines were updated to use `GinkgoRecover` so failures are reported correctly, and focused test instructions were added to make the new test easier to run.

### DIFF
--- a/.github/instructions/focused-tests.instructions.md
+++ b/.github/instructions/focused-tests.instructions.md
@@ -1,7 +1,7 @@
 ---
 name: Focused Test Execution
 description: "Use when running, debugging, or suggesting focused tests in MOCO. Covers Makefile targets, FOCUS, envtest, CI workflow/action patterns, and avoiding raw go test detours."
-applyTo: "**/*_test.go"
+applyTo: "**/*.go"
 ---
 
 # Focused test execution
--- a/.github/instructions/focused-tests.instructions.md
+++ b/.github/instructions/focused-tests.instructions.md
@@ -1,0 +1,36 @@
+---
+name: Focused Test Execution
+description: "Use when running, debugging, or suggesting focused tests in MOCO. Covers Makefile targets, FOCUS, envtest, CI workflow/action patterns, and avoiding raw go test detours."
+applyTo: "**/*_test.go"
+---
+
+# Focused test execution
+
+When running a specific MOCO test case, first use the same entry points as CI and the Makefile. Do not start with raw `go test` unless there is a clear reason.
+
+## Preferred focused commands
+
+- Controller envtest: `make controller-envtest FOCUS='part of spec name'`
+- Clustering envtest: `make clustering-envtest FOCUS='part of spec name'`
+- API envtest: `make api-envtest FOCUS='part of spec name'`
+- Backup envtest: `make backup-envtest FOCUS='part of spec name'`
+- Full envtest set: `make envtest`
+- Small package tests/install/vet/gofmt: `make test`
+- MySQL integration tests: `make test-bkop MYSQL_VERSION='8.4.8'` or `make test-dbop MYSQL_VERSION='8.4.8'`
+- E2E and upgrade tests should follow `.github/actions/e2e/action.yaml` and `.github/actions/upgrade/action.yaml`; use the `e2e/Makefile` targets from inside `e2e/`.
+
+## Why Make targets first
+
+- `Makefile` exports `ENVTEST_KUBERNETES_VERSION` and `ENVTEST_ASSETS_DIR`; raw controller/envtest commands can fail with `could not parse "v" as version` if these are missing.
+- The `*-envtest` targets run with the repo's expected flags: `-race`, `-count 1`, `-ginkgo.randomize-all`, `-ginkgo.v`, and usually `-ginkgo.fail-fast`.
+- `controller-envtest` also sets `DEBUG_CONTROLLER=1`.
+- CI calls `make lint`, `make test`, `make check-generate`, and `make envtest` after `.github/actions/setup-aqua`, so local verification should mirror those paths when practical.
+
+## If raw `go test` is necessary
+
+Only use raw `go test` for a quick diagnostic. For envtest packages, carry over the Makefile environment explicitly:
+
+```sh
+ENVTEST_KUBERNETES_VERSION=1.35.0 ENVTEST_ASSETS_DIR="$PWD/bin" \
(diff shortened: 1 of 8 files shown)


yoheinbb:
@arosh 
There are merge conflicts in this PR. Could you please resolve them?

arosh:
@yoheinbb 
resolved 78e1c9e85e96367ee53d6caaede80311c0592a23

yoheinbb (approved):

yamatcha (approved):

PR #925, opened 2026-06-19
Add predicates to PodWatcher for better event filtering

arosh:
## Summary
This change reduces unnecessary controller reconciliations by filtering out unrelated Pods earlier. It keeps the existing behavior for MOCO-managed Pods, but avoids processing Pods that do not match the expected labels.

## Context
Before this change, the controllers could receive more events than needed and then reject them during reconciliation. That added extra work and made the watch path less efficient. The goal here is to filter events as early as possible while keeping the same ownership checks in the reconcile logic.

## Implementation
`PodWatcher` now uses a predicate so it only watches Pods with the MOCO MySQL labels.

## Note
`StatefulSetPartitionReconciler` uses the same style of predicate for StatefulSets and owned Pods, replacing the older event-specific predicate code with a simpler object-level check using [NewPredicateFuncs](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.23.3/pkg/predicate#NewPredicateFuncs).

### DIFF
--- a/controllers/partition_controller.go
+++ b/controllers/partition_controller.go
@@ -21,7 +21,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
-	"sigs.k8s.io/controller-runtime/pkg/event"
 	crlog "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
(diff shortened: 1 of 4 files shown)


yamatcha (dismissed):

controllers/partition_controller.go
```diff
@@ -135,14 +133,7 @@ func partitionControllerPredicate() predicate.Funcs {
...
 		return true
-	}
-
-	return predicate.Funcs{
-		UpdateFunc:  func(e event.UpdateEvent) bool { return prctFunc(e.ObjectNew) },
-		CreateFunc:  func(e event.CreateEvent) bool { return prctFunc(e.Object) },
-		DeleteFunc:  func(e event.DeleteEvent) bool { return prctFunc(e.Object) },
-		GenericFunc: func(e event.GenericEvent) bool { return prctFunc(e.Object) },
-	}
+	})
```
yamatcha: This  also can be implemented with `predicate.LabelSelectorPredicate`.

arosh: Based on personal communication with yamatcha, it was decided to omit the check for the `moco-` prefix. Label-based filtering is sufficient.

yamatcha (approved):

PR #946, opened 2026-09-03
Fix nil pointer dereference for an offline MySQLCluster

kuenishi:
We've come across an issue where an offline MySQL cluster takes down the controller due to nil deference, in our production cluster. It looks like a bug to me and I'm sending patch hereby to discuss how it can be fixed.

## Version

0.34-0.37

## Observation

Controller repeatedly goes down into crash loop with following log

```
moco-controller-6644549779-ljvjv moco-controller {"level":"error","ts":"2026-09-02T23:49:01Z","logger":"cluster-manager.(namespace)/(cluatername)","msg":"error","operationId":"op-j6fp4","duration":0.069841086,"error":"too few pods; only 0 pods exist","stacktrace":"github.com/cybozu-go/moco/clustering.(*managerProcess).Start\n\t/work/clustering/process.go:165\ngithub.com/cybozu-go/moco/clustering.(*clusterManager).update.func1\n\t/work/clustering/manager.go:94\nsync.(*WaitGroup).Go.func1\n\t/usr/local/go/src/sync/waitgroup.go:258"}

```

followed by stack trace

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0xb8]

github.com/cybozu-go/moco/clustering.(*managerProcess).removeAnnPreventDelete
	/work/clustering/operations.go:357
github.com/cybozu-go/moco/clustering.(*managerProcess).do
	/work/clustering/process.go:194
github.com/cybozu-go/moco/clustering.(*managerProcess).Start
	/work/clustering/process.go:160
github.com/cybozu-go/moco/clustering.(*clusterManager).update.func1
	/work/clustering/manager.go:94
```

## Description

An offline MySQLCluster has no mysqld Pods. GatherStatus skips the Pod count check when spec.offline is true, so StatusSet.Pods is left as a slice of nil pointers:

```go
    if !cluster.Spec.Offline && int(cluster.Spec.Replicas) != len(pods.Items) {
        return nil, fmt.Errorf("too few pods; only %d pods exist", len(pods.Items))
    }
    ss.Pods = make([]*corev1.Pod, cluster.Spec.Replicas)
    for i, pod := range pods.Items { ... }  // no iteration when there is no Pod
```

GatherStatus still succeeds in this case: when an instance address cannot be resolved, dbop's factory returns a NopOperator rather than an error, and GetStatus then reports ErrNop, which the status loop treats as "no status" instead of a failure.

do() handles the prevent-delete annotation before the ss.State switch, so removeAnnPreventDelete dereferences one of those nil Pods and panics. The panic happens in a goroutine started by clusterManager.update, so it is not recovered and the whole moco-controller process crashes. A single offline MySQLCluster therefore takes down the controller for every MySQLCluster it manages.

Return early from do() when the cluster is offline. The switch below already handles StateOffline by returning (false, nil), so this only moves the check ahead of the code that cannot handle nil Pods.

### DIFF
--- a/clustering/process.go
+++ b/clustering/process.go
@@ -185,6 +185,15 @@ func (p *managerProcess) do(ctx context.Context) (bool, error) {
 		return false, fmt.Errorf("failed to update status fields in MySQLCluster: %w", err)
 	}
 
+	// An offline cluster has no mysqld Pods. GatherStatus skips the Pod count
+	// check when spec.offline is true, so ss.Pods is a slice of nil pointers.
+	// Handling the prevent-delete annotation would dereference them, and the
+	// StateOffline branch below does nothing anyway, so return here.
+	if ss.State == StateOffline {
+		logFromContext(ctx).Info("cluster state is " + ss.State.String())
+		return false, nil
+	}
+
 	if ss.PreventPodDeletion {
 		err := p.addAnnPreventDelete(ctx, ss)
 		if err != nil {
--- a/clustering/process_test.go
+++ b/clustering/process_test.go
@@ -1,0 +1,94 @@
+package clustering
+
+import (
+	"context"
+	"testing"
+
+	mocov1beta2 "github.com/cybozu-go/moco/api/v1beta2"
+	"github.com/cybozu-go/moco/pkg/dbop"
+	"github.com/cybozu-go/moco/pkg/metrics"
+	"github.com/cybozu-go/moco/pkg/password"
+	"github.com/prometheus/client_golang/prometheus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/tools/record"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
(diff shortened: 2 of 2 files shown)


kuenishi:
To clarify, I meant "an offline cluster" is a MySQLCluster having `offline: true` in its spec.

clustering/process.go
kuenishi, line 212: Copilot says this branch is useless, which is true. What is the best practice here in MOCO? I would recommend returning `fmt.Errorf("..")` like assertion. But it can be just failing with some fatal, or just removing this case?

yamatcha:
@kuenishi 

Thank you for the detailed report and the patch - I reproduced the panic and the analysis is correct.

One problem with the early return is that `prevent-delete` is enforced by a ValidatingAdmissionPolicy that denies Pod DELETE. If a cluster goes offline while any of its Pods still has the annotation (normally the primary), `removeAnnPreventDelete` never runs again, so the StatefulSet scale-down to zero stays blocked forever.

Could you drop the `do()` change and instead make the two helpers nil-safe?
```
    // removeAnnPreventDelete
    for _, pod := range ss.Pods {
        if pod == nil { // Pods of an offline cluster may not exist
            continue
        }

    // addAnnPreventDelete
    ppod := ss.Pods[ss.Primary]
    if ppod == nil {
        return nil
    }
```

This fixes the reported panic plus a second potential panic in `addAnnPreventDelete`, keeps the annotation cleanup working during the offline scale-down, and your test passes unchanged.

